Issue #4, opened 2023-09-27
[CLI] Implement the `routes` console task


rsamoilov:
Enable users to examine available routes by running `rage routes`. The output should be pretty much similar to `rails routes` and include the `Verb`, `URI Pattern`, and `Controller#Action` fields.

💡 The change should be made in [lib/rage/cli.rb](https://github.com/rage-rb/rage/blob/master/lib/rage/cli.rb).

PR #9, opened 2023-10-05
Add routes cli command

arikarim:
Resolves[issue](https://github.com/rage-rb/rage/issues/4) by adding a simple routes command to rage cli.

### DIFF
--- a/lib/rage/cli.rb
+++ b/lib/rage/cli.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
-
-require 'thor'
-require 'rage'
+require "thor"
+require "rage"
 
 module Rage
   class CLI < Thor
@@ -9,14 +8,14 @@ module Rage
       true
     end
 
-    desc 'new PATH', 'Create a new application.'
+    desc "new PATH", "Create a new application."
     def new(path)
       NewAppGenerator.start([path])
     end
 
-    desc 's', 'Start the app server.'
+    desc "s", "Start the app server."
     def server
-      app = ::Rack::Builder.parse_file('config.ru')
+      app = ::Rack::Builder.parse_file("config.ru")
       app = app[0] if app.is_a?(Array)
 
       ::Iodine.listen service: :http, handler: app, port: Rage.config.port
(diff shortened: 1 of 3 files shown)


lib/rage/cli.rb
```diff
@@ -25,6 +25,40 @@ def server
...
+    def routes
+      # the result would be something like this:
+      # +----------------------+----------------------+----------------------+
+      # | Action               | Verb                 | Path                 | Controller#Action
+      # | index                | GET                  | /                    | application#index
+
+      # load config/application.rb
+      require File.expand_path('config/application.rb', Dir.pwd)
+      # load config/routes.rb
+      require File.expand_path('config/routes.rb', Dir.pwd)
```
rsamoilov, line 39: you only need to load the `config/application` file, as it [loads all routes automatically](https://github.com/rage-rb/rage/blob/master/lib/rage/setup.rb). Also, let's move this require call into a method, as this is the only task here that needs access to the application code.

lib/rage/cli.rb
```diff
@@ -25,6 +25,40 @@ def server
...
+
+      routes = Rage.routes.router.routes
+
+      # construct a table
+      table = [['+---------------------------', '+---------------------------', '+---------------------------+']]
+
+      table << ['| Action'.ljust(20), '| Verb'.ljust(20), '| Path'.ljust(20), "| Controller#Action\n\n"]
```
rsamoilov, line 46: We should find the longest verb (that is always `DELETE`) and the longest path, and align based on those values, as there might be paths longer than 20 characters. I would suggest to use `sprintf` as it might work good here.

lib/rage/cli.rb
```diff
@@ -25,6 +25,40 @@ def server
...
+
+      # load config/application.rb
+      require File.expand_path('config/application.rb', Dir.pwd)
+      # load config/routes.rb
+      require File.expand_path('config/routes.rb', Dir.pwd)
+
+      routes = Rage.routes.router.routes
+
+      # construct a table
+      table = [['+---------------------------', '+---------------------------', '+---------------------------+']]
```
rsamoilov, line 44: Let's try to mimic the output of `rails routes` but without the first `Prefix` column. That would also mean removing all table formatting.

rsamoilov (commented):
Hey Ari, I've left some comments.
Thank you for the PR!

arikarim:
Hello @rsamoilov, Thanks for the beautiful comments.

I have now made the changes you asked.

- [ ] Use of ` Rage.routes.router.routes`
- [ ]  Using format instead of sprintf per rubocop recommendation and according to the longest string.

lib/rage/router/dsl.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Rage::Router::DSL
+  attr_reader :router
```
rsamoilov, line 4: This is not needed anymore.

rsamoilov (commented):
Hey @arikarim 

This looks good!

Some more comments:

- The task currently fails when there are no routes.
- The task currently fails when there are lambda routes, e.g. https://github.com/rage-rb/rage/blob/master/lib/rage/templates/config-routes.rb.
- The task attempts to create a command based on the `require_file_by_path` method. Can you please move this method inside a `no_commands` block?
- Can we have a consistent margin of three spaces between the columns? Currently margins are inconsistent, see the screenshot.
- Can we have just one line break after the header? Currently there are two, see the screenshot.
- Can we also add constraints and defaults to the output the same way it works in Rails?
- The last one: do you mind reverting the changes outside the `routes` method, like [this](https://github.com/rage-rb/rage/pull/9/files#diff-0a4e8ad14df9dde28e4e400e0c67231b39fdd7b22c0fa7f2d703c87bd748d6b7L3-R4)?

arikarim:
for the 3 space alignment there's a small issue which is that the header sometimes is longer which makes the body rows to move left as the screenshot below:
image

so the solution is to make them aligned to at least 3 space like:
image

arikarim:
I have done most of the changes you asked @rsamoilov, thanks for your comments, 
i  moved the `require_file_by_path` under private so thor does not treat it as a command
for lambdas i just added `Lambda` string instead of their name
I'm just not sure what do you mean by `Can we also add constraints and defaults to the output the same way it works in Rails?`

arikarim:
Also I have added `grep` option as in rails

rsamoilov:
This looks fantastic! Thank you!

Regarding the `defaults` and `constraints` options:
In Rails, if you define a route with defaults or constraints, `rails routes` will also display this info.
For example:

```
get "comments/:id/:format", to: "application#index", defaults: { id: -1 }, constraints: { format: /markdown/ }
```

will display the following data in the `Controller#Action` column for this route:

```
{:id=>-1, :format=>/markdown/}
```

If you are feeling like it, feel free to tackle this in another PR.

arikarim:
Thanks, sure i will soon.:)

PR #14, opened 2023-10-09
Implement match router method

arikarim:


arikarim:
Hey @rsamoilov :).
I have implemented the basics of match route with a couple of tests, it seems to work fine for various cases.

Please take a look and let me know whats need to be changed.
Thanks

rsamoilov (commented):
Hey @arikarim 

So, you are moving in the right direction, but `via` is not really a constraint. Which means you don't need to add another strategy.

`via` just specifies which methods to add for a route.

A couple examples:

```ruby
match "/photos/:id", to: "photos#show", via: :all # adds all five routes
match "/photos/:id", to: "photos#show" # the same as `via: :all`
match "/photos/:id", to: "photos#show", via: :get # adds just the GET
match "/photos/:id", to: "photos#show", via: [:post, :put] # adds only POST and PUT
```

As you can see, `via` can be either a symbol or an array of symbols. And if `via` is either `nil` or `:all` means all default methods should be used.

arikarim:
Thanku @rsamoilov 
I did what you suggested, and it seems to work perfectly :)

lib/rage/router/dsl.rb
```diff
@@ -98,6 +98,33 @@ def root(to:)
...
+      # via is either empty, or an array of strings
+      via = [via].flatten.compact.reject(&:empty?).map(&:to_s)
+
+      if via.any? && !via.include?('all')
+        via.flatten.each do |method|
+          send(method.downcase, path, to: to, constraints: constraints, defaults: defaults)
+        end
+      else
+        # if via is 'all' or empty, then we get all possible methods except 'all' and add them
+        allowed_methods = %w[get post put patch delete]
```
rsamoilov, line 121: Let's extract this into a constant.

lib/rage/router/dsl.rb
```diff
@@ -98,6 +98,33 @@ def root(to:)
...
+    # @param path [String] the path for the route handler
+    # @param to [String] the route handler in the format of "controller#action"
+    # @param constraints [Hash] a hash of constraints for the route
+    # @param defaults [Hash] a hash of default parameters for the route
+    # @param via [Symbol, Array<Symbol>] an array of HTTP methods to accept
+    # @example
+    #   match "/photos/:id", to: "photos#show", via: ["get", "post"]
+    # @example
+    #   match "/photos/:id", to: "photos#show", via: :all
+    def match(path, to:, constraints: {}, defaults: nil, via: nil)
```
rsamoilov, line 111: Seems to make more sense to initialize `via` as `all`.

```suggestion
    def match(path, to:, constraints: {}, defaults: nil, via: :all)
```

lib/rage/router/dsl.rb
```diff
@@ -98,6 +98,33 @@ def root(to:)
...
+      # via is either empty, or an array of strings
+      via = [via].flatten.compact.reject(&:empty?).map(&:to_s)
```
rsamoilov, line 113: Can we simplify this a bit? E.g.

```ruby
via = DEFAULT_MATCH_METHODS if via == :all
Array(via).each do
   ...
end
```

spec/router/dsl_spec.rb
```diff
@@ -186,4 +186,39 @@
...
+      expect(router).to receive(:on).with("POST", "/test", "test#index", constraints: {}, defaults: nil)
+      expect(router).to receive(:on).with("PUT", "/test", "test#index", constraints: {}, defaults: nil)
+      expect(router).to receive(:on).with("PATCH", "/test", "test#index", constraints: {}, defaults: nil)
+      expect(router).to receive(:on).with("DELETE", "/test", "test#index", constraints: {}, defaults: nil)
+
+      dsl.draw do
+        match "/test", to: "test#index"
+      end
+    end
+  end
```
rsamoilov, line 223: I love the tests!

Can you please also add one to check if constraints and defaults are correctly passed down? And another one to check if `match` works correctly inside `scope`.

rsamoilov (commented):
Hey @arikarim 

Thanks for another awesome PR.

rsamoilov (approved):
Looks good!

PR #15, opened 2023-10-10
Add tests for new header expose

### LINKED ISSUE #3: [Controller] Allow to access request headers
Expose a `request` object with the `headers` property to allow users to access request headers. Users should be able to access headers data by both meta-variable names (e.g. `HTTP_CONTENT_TYPE`) and original names (e.g. `Content-Type`).

Example:

```ruby
request.headers["Content-Type"]
```

Make sure to thoroughly cover the change with tests and add documentation using [YARD](https://rubydoc.info/gems/yard/file/docs/Tags.md). Also, please don't use the `Rack::Request` object.

💡 The change should be made in [rage/controller/api.rb](https://github.com/rage-rb/rage/blob/master/lib/rage/controller/api.rb).
💡 Use `yard server --reload` to verify your documentation changes.

❗ Do not hesitate to use [benchmark-ips](https://github.com/evanphx/benchmark-ips) whenever there are different implementation options. When working on gems, performance is much more important than readability.

***Please comment on this issue or assign to yourself in case you have started to work on it.***

arikarim:
Resolves #3

### DIFF
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -214,11 +214,13 @@ class RageController::API
   DEFAULT_HEADERS = { "content-type" => "application/json; charset=utf-8" }.freeze
 
   # @private
+  attr_reader :request
   def initialize(env, params)
     @__env = env
     @__params = params
     @__status, @__headers, @__body = 204, DEFAULT_HEADERS, []
     @__rendered = false
+    @request = Request.new(env)
   end
 
   # Send a response to the client.
@@ -282,5 +284,34 @@ class RageController::API
     # copy-on-write implementation for the headers object
     @__headers = {}.merge!(@__headers) if DEFAULT_HEADERS.equal?(@__headers)
     @__headers
+  end
+
+  class Request
+    # Get the request headers.
+    # @example
+    #  request.headers["Content-Type"] # => "application/json"
+    # or request.headers["HTTP_CONTENT_TYPE"] # => "application/json"
+    attr_reader :headers
+
+    def initialize(env)
+      @env = env
+      @headers = extract_headers(env)
+    end
+
+    private
+
+    def extract_headers(env)
+      headers_hash = {}
+
+      env.each do |key, value|
+        if key.start_with?('HTTP_')
+          original_name = key.sub('HTTP_', '').split('_').map(&:capitalize).join('-')
+          headers_hash[original_name] = value
+          headers_hash[key] = value
+        end
+      end
+
+      headers_hash
+    end
   end
 end
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -167,7 +167,7 @@ class RageController::API
         if: _if,
         unless: _unless
       }
-      
+
       action[:if] = define_tmp_method(action[:if]) if action[:if].is_a?(Proc)
       action[:unless] = define_tmp_method(action[:unless]) if action[:unless].is_a?(Proc)
 
@@ -293,8 +293,6 @@ class RageController::API
     # @example
     #  request.headers["Content-Type"] # => "application/json"
     # or request.headers["HTTP_CONTENT_TYPE"] # => "application/json"
-    attr_reader :headers
-
     def initialize(env)
       @env = env
     end
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -214,7 +214,6 @@ class RageController::API
   DEFAULT_HEADERS = { "content-type" => "application/json; charset=utf-8" }.freeze
 
   # @private
-  attr_reader :request
   def initialize(env, params)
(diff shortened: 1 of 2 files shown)


arikarim:
@rsamoilov sorry for the bothering :)

lib/rage/controller/api.rb
```diff
@@ -283,4 +285,33 @@ def headers
     @__headers = {}.merge!(@__headers) if DEFAULT_HEADERS.equal?(@__headers)
     @__headers
   end
+
+  class Request
```
rsamoilov, line 289: Unfortunately, it's a bit more complicated.

We can't iterate through the whole env normalizing header names, as it is inefficient. A user might only need one header value, but your implementation will loop through all of them to do that.

On top of that, the `Content-Type` and `Content-Length` headers are not getting the `HTTP_` prefix.

Instead, nothing should be done until an actual header is requested. When it is, the code should look something like this:

```ruby
if requested_header.start_with?("HTTP_")
  env[requested_header]
else
  normalized_name = ... # normalize the header name, e.g. User-Agent -> HTTP_USER_AGENT
  env[normalized_name]
end
```

Also, the code should somehow make exceptions for the two Content headers.

arikarim:
Hey @rsamoilov  I did what I could, please feel free to close the pull request if you felt that you need another approach.

lib/rage/controller/api.rb
```diff
@@ -283,4 +288,42 @@ def headers
...
+    end
+
+    class Headers
+      def initialize(env)
+        @env = env
+        @headers = {}
+      end
+
+      def [](requested_header)
+        requested_header = requested_header.upcase.tr('-', '_')
```
rsamoilov, line 315: so I've run this benchmark:

```ruby
require "benchmark/ips"

requested_header = "Content-Type".freeze

Benchmark.ips do |x|
  x.report("1") { requested_header.upcase.tr!("-", "_") }
  x.report("2") { s = requested_header.tr("-", "_"); s.upcase! || s }

  x.compare!
end && true
```

and received the following result:

```
Warming up --------------------------------------
                   1   308.241k i/100ms
                   2   442.991k i/100ms
Calculating -------------------------------------
                   1      3.043M (± 1.5%) i/s -     15.412M in   5.066559s
                   2      4.575M (± 0.5%) i/s -     23.036M in   5.035321s

Comparison:
                   2:  4574912.3 i/s
                   1:  3042613.1 i/s - 1.50x  slower
```

Based on this, let's move this line inside the `else` branch and change it accordingly.

lib/rage/controller/api.rb
```diff
@@ -283,4 +288,42 @@ def headers
...
+
+        if requested_header.start_with?("HTTP_")
+          @env[requested_header]
+        else
+          normalized_name = "HTTP_" + requested_header
+          if SPECIAL_HEADERS.include?(requested_header)
+            normalized_name = requested_header
+          end
+          @env[normalized_name]
```
rsamoilov, line 324: again, for performance reasons, let's update this:

```suggestion
          if "CONTENT_TYPE" == requested_header || "CONTENT_LENGTH" == requested_header
            @env[requested_header]
          else
            (requested_header = requested_header.tr('-', '_')).upcase!
            @env["HTTP_#{requested_header}"]
          end
```

lib/rage/controller/api.rb
```diff
@@ -283,4 +288,42 @@ def headers
...
+    end
+
+    def headers
+      @headers ||= Headers.new(@env)
+    end
+
+    class Headers
+      def initialize(env)
+        @env = env
+        @headers = {}
```
rsamoilov, line 311: This variable seems to be not needed.

lib/rage/controller/api.rb
```diff
@@ -283,4 +288,42 @@ def headers
     @__headers = {}.merge!(@__headers) if DEFAULT_HEADERS.equal?(@__headers)
     @__headers
   end
+
+  class Request
+    SPECIAL_HEADERS = %w[CONTENT_TYPE CONTENT_LENGTH].freeze
```
rsamoilov, line 293: Not needed.

lib/rage/controller/api.rb
```diff
@@ -214,13 +214,18 @@ def skip_before_action(action_name, only: nil, except: nil)
   DEFAULT_HEADERS = { "content-type" => "application/json; charset=utf-8" }.freeze
 
   # @private
+  attr_reader :request
```
rsamoilov, line 217: Not needed.

rsamoilov (commented):
This is exactly what I've had in my mind! Thank you @arikarim .

I've left some comments, but it's already almost perfect.

arikarim:
Thanku for the comments, It should be good now

rsamoilov (approved):
Merged! Thank you!

PR #18, opened 2023-10-12
Add mount method

arikarim:
Add mount method

### DIFF
--- a/lib/rage/router/dsl.rb
+++ b/lib/rage/router/dsl.rb
@@ -175,6 +175,15 @@ class Rage::Router::DSL
       @defaults.pop
     end
 
+    # Mount a Rack-based application to be used within the application.
+    #
+    # @param app [Rack::Application] a Rack-based application
+    # @param at [String] the path at which to mount the Rack-based application
+    # @param via [Symbol] the HTTP method to use when mounting the Rack-based application
+    # @example
+    #   mount Sidekiq::Web => '/sidekiq', via: :get
+    # or
+    #  mount Sidekiq::Web, at: '/sidekiq', via: :get
     def mount(*args)
       if args.first.is_a?(Hash)
         app = args.first.keys.first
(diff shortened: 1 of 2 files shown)


rsamoilov:
Hey @arikarim 

I'm sorry I missed your PRs! I will make sure to review these today. Thank you!

arikarim:
No worries, Thanks

lib/rage/router/dsl.rb
```diff
@@ -175,6 +175,24 @@ def defaults(defaults, &block)
...
+        at = args.first.values.first
+        via = args[0][:via]
+      else
+        app = args.first
+        at = args[1][:at]
+        via = args[1][:via]
+      end
+
+      # If app is a class, convert it to a lambda
+      app = ->(env) { app.new.call(env) } if app.is_a?(Class)
```
rsamoilov, line 190: It is expected that a Rack application responds to `call`, so I believe this is not needed.

lib/rage/router/dsl.rb
```diff
@@ -175,6 +175,24 @@ def defaults(defaults, &block)
...
+
+      # Use match with via: :all to mount the Rack-based application
+      match(at, to: app, via:)
```
rsamoilov, line 193: Should it be something like `"#{at}/*"` in order to match all routes of the Rack application?

spec/router/dsl_spec.rb
```diff
@@ -264,4 +264,28 @@
       end
     end
   end
+
+  context "Mount" do
```
rsamoilov, line 268: Some tests are missing:

* for the hash version of `mount`;
* without the `via` option;
* with a class instead of a lambda;

lib/rage/router/dsl.rb
```diff
@@ -175,6 +175,24 @@ def defaults(defaults, &block)
       @defaults.pop
     end
 
+    def mount(*args)
```
rsamoilov, line 178: Can you please add documentation?

lib/rage/router/dsl.rb
arikarim: good idea

arikarim: I don't think so, if the user want to do that he just can say:   mount Sidekiq::Web =>` '/sidekiq/*'`

arikarim:
Thanku, I have done the changes.

rsamoilov (approved):
👍

PR #50, opened 2023-12-12
Autoloading and reloading with Zeitwerk

alex-rogachev:
Added Zeitwerk and Filewatcher dependencies, initial logic to autoload app directory

### DIFF
--- a/lib/rage/code_loader.rb
+++ b/lib/rage/code_loader.rb
@@ -10,7 +10,7 @@ class Rage::CodeLoader
   def setup
     autoload_path = "#{Rage.root}/app"
     enable_reloading = Rage.env.development?
-    enable_eager_loading = Rage.env.production?
+    enable_eager_loading = !Rage.env.development? && !Rage.env.test?
 
     @loader.push_dir(autoload_path)
     # The first level of directories in app directory won't be treated as modules
--- a/lib/rage/code_loader.rb
+++ b/lib/rage/code_loader.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require "zeitwerk"
+
+class Rage::CodeLoader
+  def initialize
+    @loader = Zeitwerk::Loader.new
+  end
+
+  def setup
+    autoload_path = "#{Rage.root}/app"
+    enable_reloading = Rage.env.development?
+    enable_eager_loading = Rage.env.production?
+
+    @loader.push_dir(autoload_path)
+    # The first level of directories in app directory won't be treated as modules
+    # e.g. app/controllers/pages_controller.rb will be linked to PagesController class
+    # instead of Controllers::PagesController
+    @loader.collapse("#{Rage.root}/app/*")
+    @loader.enable_reloading if enable_reloading
+    @loader.setup
+    @loader.eager_load if enable_eager_loading
+  end
+
+  def reload
+    @loader.reload
+    Rage.__router.reset_routes
+    load("#{Rage.root}/config/routes.rb")
+  end
+
+  def reloading_enabled?
+    @loader.reloading_enabled?
+  end
+end
--- a/lib/rage/middleware/reloader.rb
+++ b/lib/rage/middleware/reloader.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class Rage::Reloader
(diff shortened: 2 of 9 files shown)


lib/rage/code_loader.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def reload
+    @loader.reload
+    Rage.__router.reset_routes
+    load("#{Rage.root}/config/routes.rb")
+  end
+
+  def reloading_enabled?
+    @loader.reloading_enabled?
```
rsamoilov, line 32: If I'm getting this right, this method always returns `true` if `enable_reloading` has been called. This means some of the errors we purposely ignore on reloading will also be ignored on server startup. Which will lead to different behaviour in development and production.

Let's use a custom flag here.

lib/rage/middleware/reloader.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class Rage::Reloader
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    Rage.code_loader.reload
+    @app.call(env)
```
rsamoilov, line 10: Let's also catch errors here.

```suggestion
    @app.call(env)
rescue Exception => e
    exception_str = "#{e.class} (#{e.message}):\n#{e.backtrace.join("\n")}"
    puts(exception_str)
    [500, {}, [exception_str]]
```

lib/rage/router/backend.rb
```diff
@@ -63,6 +68,7 @@ def on(method, path, handler, constraints: {}, defaults: nil)
       raise "Invalid route handler format, expected to match the 'controller#action' pattern" unless handler =~ STRING_HANDLER_REGEXP
 
       controller, action = to_controller_class($1), $2
+      return if controller.nil? && Rage.code_loader.reloading_enabled?
       run_action_method_name = controller.__register_action(action.to_sym)
```
rsamoilov, line 72: It looks like we also need another `return` after this line to prevent the router from creating an incorrect handler.

lib/rage/templates/config-environments-development.rb
```diff
@@ -7,4 +7,6 @@
 
   # Specify the logger
   config.logger = Rage::Logger.new(STDOUT)
+
+  config.middleware.use Rage::Reloader
```
rsamoilov, line 11: 😍

lib/rage/code_loader.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class Rage::CodeLoader
+  def initialize
+    @loader = Zeitwerk::Loader.new
+  end
+
+  def setup
+    autoload_path = "#{Rage.root}/app"
+    enable_reloading = Rage.env.development?
+    enable_eager_loading = Rage.env.production?
```
rsamoilov, line 13: What do you think about changing the condition here?

```suggestion
    enable_eager_loading = !Rage.env.development? && !Rage.env.test?
```

rsamoilov (commented):
This looks awesome!

Here's an idea I had - what if we update all exceptions we need to ignore on reloading with a custom class, e.g. `Rage::Error`? Then, we will only need one `return if Rage.code_loader.reloading_enabled?` condition somewhere in `Rage::Router::Backend#on` to ignore all of them at once. What do you think?

lib/rage/code_loader.rb
alex-rogachev: Totally makes sense!

lib/rage/middleware/reloader.rb
alex-rogachev: This was so obvious to catch all exceptions in the middleware itself 🙈 Shame I didn't see this solution but it is really neat!

lib/rage/code_loader.rb
alex-rogachev: That's true. Will do!

alex-rogachev: @rsamoilov wait... Do we even need this flag now? The application will try to draw routes on server startup (in setup.rb) so it will fail if, for example some controller action is missing. And during request the middleware will catch all exception and present it to the end user. I don't really see why we need the reloading flag now.

PR #54, opened 2023-12-18
Zeitwerk autoloading and reloading

alex-rogachev:


rsamoilov (commented):
let's also ignore [this](https://github.com/rage-rb/rage/blob/8a82d4fe7f3aa45638a0ba3a055c8c49fa43d1dd/lib/rage/controller/api.rb#L192) error on reloading.

rsamoilov (approved):
This is awesome!

PR #55, opened 2023-12-21
Add conditional GET using `stale?`

tonekk:
This PR adds conditional GET, that is [planned for version 0.8](https://github.com/rage-rb/rage#upcoming-releases).

## Some background info
While the goal of rage is to stay compliant to the Rails API, we decided to slightly diverge from it here.
Rails also has `fresh_when`, which in our case does no really make sense:
1. It's first argument (`object`) needs to respond to `updated_at`, which would create tight coupling to an ActiveRecord(ish) library, which might not be what we want
2. `fresh_when` is only practical when an implicit render happens afterwards. Rails API controllers do not support implicit rendering and neither does rage. Explicitly calling render leads to a double render error.

This is why we decided to stick to `stale?` exclusively:
- It does not take an `object` as an argument
- It can be used to guard against double renders:
  ```ruby
    def index
      return unless stale?(etag: "foo")

      render @items
    end
  ```

Furthermore, we decided that ETag generation will not be part of rage for now. As we gather some data of people actually using rage in production, we will be able to see if and how this feature is used. For now, all we have is assumptions and copying Rails' behaviour seems unreasonable here, since it again relies on ActiveRecord (`cache_key`, in this case).

### To be discussed

* Since the user has to generate ETags for themselves, I was wondering if having multiple parameters (`etag`, `strong_etag` and `weak_etag`) was still justifiable. Not to mention that, in order to __really__ support strong ETags, we would have to implement a byte-by-byte comparison (i.e. rendering the action, creating a digest and compare that). [Even Rails doesn't do it](https://www.bigbinary.com/blog/rails-5-switches-from-strong-etags-to-weak-tags).
* Do we want the `public` parameter to control the publicness of `Cache-Control` headers? I wasn't sure if that might be out of scope, so I left it out for now.

### Additional notes

* I also removed the `template` parameter, because we're not using templates.

### DIFF
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -331,4 +331,26 @@ class RageController::API
       @params ||= ActionController::Parameters.new(@__params)
     end
   end
+
+  # Checks if the request is stale to decide if the action has to be rendered or the cached version is still valid. Use this method to implement conditional GET.
+  #
+  # @param etag [String] The etag of the requested resource.
+  # @param last_modified [Time] The last modified time of the requested resource.
+  # @return [Boolean] True if the response is stale, false otherwise.
+  # @example
+  #  stale?(etag: "123", last_modified: Time.utc(2023, 12, 15))
+  #  stale?(last_modified: Time.utc(2023, 12, 15))
+  #  stale?(etag: "123")
+  # @note `stale?` will set the response status to 304 if the request is fresh. This side effect will cause a double render error, if `render` gets called after this method. Make sure to implement a proper conditional in your action to prevent this from happening:
+  #  ```ruby
+  #  if stale?(etag: "123")
+  #    render json: { hello: "world" }
+  #  end
+  #  ```
+  def stale?(etag: nil, last_modified: nil)
+    still_fresh = request.fresh?(etag:, last_modified:)
+
+    head :not_modified if still_fresh
+    !still_fresh
+  end
 end
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -14,6 +14,48 @@ class Rage::Request
     @headers ||= Headers.new(@env)
   end
 
+  # Check if the request is fresh.
+  # @param etag [String] The etag of the requested resource.
+  # @param last_modified [Time] The last modified time of the requested resource.
+  # @return [Boolean] True if the request is fresh, false otherwise.
+  # @example
+  #  request.fresh?(etag: "123", last_modified: Time.utc(2023, 12, 15))
+  #  request.fresh?(last_modified: Time.utc(2023, 12, 15))
+  #  request.fresh?(etag: "123")
+  def fresh?(etag:, last_modified:)
+    if_none_match = headers["HTTP_IF_NONE_MATCH"]
+    if_not_modified_since = headers["HTTP_IF_MODIFIED_SINCE"]
+
+    # Always render response when no freshness information
+    # is provided in the request.
+    return false unless if_none_match || if_not_modified_since
+
+    etag_matches?(
+      requested_etags: if_none_match, response_etag: etag
+    ) && not_modified?(
+      request_not_modified_since: if_not_modified_since,
+      response_last_modified: last_modified
+    )
+  end
+
+  private
+
+  def etag_matches?(requested_etags:, response_etag:)
+    requested_etags = requested_etags ? requested_etags.split(",").map(&:strip) : []
+
+    return true if requested_etags.empty?
+    return false if response_etag.nil?
+
+    requested_etags.include?(response_etag) || requested_etags.include?("*")
+  end
+
+  def not_modified?(request_not_modified_since:, response_last_modified:)
+    return true if request_not_modified_since.nil?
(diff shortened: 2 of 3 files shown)


lib/rage/request.rb
```diff
@@ -14,6 +14,48 @@ def headers
...
+    ) && not_modified?(
+      request_not_modified_since: if_not_modified_since,
+      response_last_modified: last_modified
+    )
+  end
+
+  private
+
+  def etag_matches?(requested_etags:, response_etag:)
+    requested_etags = requested_etags ? requested_etags.split(",").map(&:strip) : []
```
rsamoilov, line 44:
```suggestion
    requested_etags = requested_etags ? requested_etags.split(",").map! { |tag| tag.strip! || tag } : []
```

rsamoilov: Also, do we have to create an empty array if there are no `requested_etags`? Can we go with `nil` instead?

lib/rage/request.rb
```diff
@@ -14,6 +14,48 @@ def headers
...
+
+    return true if requested_etags.empty?
+    return false if response_etag.nil?
+
+    requested_etags.include?(response_etag) || requested_etags.include?("*")
```
rsamoilov, line 49: Genuinely surprised this is much faster than calling `any?` 😯

rsamoilov:
🔥🔥🔥

lib/rage/controller/api.rb
```diff
@@ -331,4 +331,26 @@ def params
       @params ||= ActionController::Parameters.new(@__params)
     end
   end
+
+  # Checks if the request is stale to decide if the action has to be rendered or the cached version is still valid. Use this method to implement conditional GET.
+  #
+  # @param etag [String] The etag of the requested resource.
+  # @param last_modified [Time] The last modified time of the requested resource.
```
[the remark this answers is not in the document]
tonekk: Why do you think `String` is more reasonable here?
If we want to do what Rails does, I think we should [stick to time](https://apidock.com/rails/ActionController/ConditionalGet/fresh_when).

On another note, this brought to my attention that I definitely need to convert to `Time` [here](https://github.com/rage-rb/rage/pull/55/files#diff-3e7ad6582a85dbe56681557a219a4c3f96181cf138ad2b169d9455b30160c9d3R27) and that the tests should probably pass strings as HTTP headers to reflect the real world.

lib/rage/request.rb
tonekk: The thing here is that
Screenshot 2023-12-22 at 10 25 06

which means that leaving it this way would mean that an empty string would be the same as not sending the header. Not sure if that's the behaviour we're looking for 🤔 
I will check how rails does it.

Also:
Using `each` instead of `map!` here is even faster and also makes more sense IMO, because we're already mutating the strings in place, so need to mutate the array at all.

lib/rage/request.rb
tonekk: Just checked, the difference is indeed remarkable!
Hail to the ruby interpreter optimization magic 💯

lib/rage/controller/api.rb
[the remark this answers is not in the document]
rsamoilov: Right, I think `Time` makes more sense here, but [this line](https://github.com/rage-rb/rage/pull/55/files#diff-3e7ad6582a85dbe56681557a219a4c3f96181cf138ad2b169d9455b30160c9d3R56) will fail if you use `Time` with the current implementation.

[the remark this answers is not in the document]
tonekk: > Right, I think Time makes more sense here, but [this line](https://github.com/rage-rb/rage/pull/55/files#diff-3e7ad6582a85dbe56681557a219a4c3f96181cf138ad2b169d9455b30160c9d3R56) will fail if you use Time with the current implementation.

Not sure what you're referring to here.
[b9c74ad](https://github.com/rage-rb/rage/pull/55/commits/b9c74ad161160d1034e124b1bbc1337bae3cfe07) now takes into account the fact that HTTP headers are always strings and converts it to `Time`. If the user is not passing a wrong type to `stale?`, the line you mentioned should work.
We could think about adding a type conversion to `stale?`, something like:
```ruby
def stale?(etag:, last_modified:)
  last_modified = Time.new(last_modified.to_s)
  # ...
end
```

This would allow the user to pass in both `String` and `Time` objects, which would add more flexibility but also bloat the code with conversion logic.
What do you think?

lib/rage/request.rb
tonekk: [ba6b43c](https://github.com/rage-rb/rage/pull/55/commits/ba6b43ccb214862ce47a4e7cbbf5fe7e74786f45)

tonekk:
Also added better specs for testing etags with whitespace and commas.

Should I squash everything before merging?

lib/rage/controller/api.rb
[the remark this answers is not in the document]
rsamoilov: I think allowing just `Time` is sufficient.

rsamoilov (dismissed):
Nice! We'll get it merged into v1.

PR #83, opened 2024-05-03
Allow to have both Rails and Rage controllers in one application

rsamoilov:


efibootmgr:
I came across a problem actually, because RageController doesn't support any gems that patch Rails controller. Like serializer gems that allow you to write something like
render json: Model, serializer: ModelSerializer
The RageController does not get those features.

rsamoilov:
huh, that's an interesting point. Are you talking about ActiveModelSerializers?

In general, given how Rage and Rails are different internally, I am not sure it would be possible to patch Rails controllers. My expectation is that gem maintainers would need to update their code to patch Rage controllers too.

efibootmgr:
Thanks for reply, i think so. It's not critical, experienced developer can add features like that manually to rage controllers of course, if really needed. Just wanted to point to one downside/problem of using Rage and Rails together.
I was also having problems with **rage console** command, when triyng to use it in rage + rails template. Using any sql queries results in
`No connection pool for 'ActiveRecord::Base' found. (ActiveRecord::ConnectionNotEstablished)`
The **rails console** work properly and gets active record connection from config.

rsamoilov:
Thanks! This should be fixed in #84.

PR #89, opened 2024-06-19
Controller parameters wrapping feature

alex-rogachev:
Implementation of wrappers for controller parameters. 

Controllers got new class method `wrap_parameters` that accepts 2 parameters: `key`(mandatory) and `options`(optional). If called, all parameters are wrapped into a nested hash like this:
`{query_param: 'queryparam', body_param: 'bodyparam', action: 'action', controller: 'controller', wrapping_key: {query_param: 'queryparam', body_param: 'bodyparam', action: 'action', controller: 'controller'}}` where `wrapping_key` is the key set with the `wrap_parameters` method.

Parameters get wrapped only if there is no param that is named like the key passed during wrap_parameters call. Also, the 'Content_Type' header has to be present for the parameters to be wrapped.

Currently 2 options are available: include and exclude. When called with _include_ option, the logic will only wrap parameters that are passed along with this option. If called with _exclude_ option, all parameters except those that were passed with the option will be wrapped.

Possible enhancements: 
- wrap only request body params
- add request_type options

### DIFF
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -77,6 +77,22 @@ class RageController::API
 
       activerecord_loaded = defined?(::ActiveRecord)
 
+      wrap_parameters_chunk = if __wrap_parameters_key
+        <<~RUBY
+          options = self.class.__wrap_parameters_options
+          
+          wrapped_params = if options[:include]
+            @__params.slice(*[options[:include]].flatten)
+          elsif options[:exclude]
+            @__params.except(*[options[:exclude]].flatten)
+          else
+            @__params
+          end
+          
+          @__params = {self.class.__wrap_parameters_key => wrapped_params}
+        RUBY
+      end
+
       class_eval <<~RUBY,  __FILE__, __LINE__ + 1
         def __run_#{action}
           #{if activerecord_loaded
@@ -85,6 +101,7 @@ class RageController::API
             RUBY
           end}
 
+          #{wrap_parameters_chunk}
           #{before_actions_chunk}
           #{action}
 
@@ -122,6 +139,7 @@ class RageController::API
 
     # @private
     attr_writer :__before_actions, :__after_actions, :__rescue_handlers
+    attr_accessor :__wrap_parameters_key, :__wrap_parameters_options
 
     # @private
     # pass the variable down to the child; the child will continue to use it until changes need to be made;
@@ -130,6 +148,8 @@ class RageController::API
       klass.__before_actions = @__before_actions.freeze
       klass.__after_actions = @__after_actions.freeze
       klass.__rescue_handlers = @__rescue_handlers.freeze
+      klass.__wrap_parameters_key = __wrap_parameters_key
+      klass.__wrap_parameters_options = __wrap_parameters_options
     end
 
     # @private
@@ -275,6 +295,20 @@ class RageController::API
       end
 
       @__before_actions[i] = action
+    end
+
+    # Initialize controller params wrapping into a nested hash.
+    # If initialized, params wrapping logic will be added to the controller.
+    #
+    # @param key [Symbol] key that the params hash will nested under
+    # @param options [Hash] wrapping options
+    # @example
+    #   wrap_parameters :user, include: [:name, :age]
+    # @example
+    #   wrap_parameters :user, exclude: :address
+    def wrap_parameters(key, options = {})
+      @__wrap_parameters_key = key
+      @__wrap_parameters_options = options
     end
 
     private
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -79,17 +79,20 @@ class RageController::API
 
       wrap_parameters_chunk = if __wrap_parameters_key
         <<~RUBY
-          options = self.class.__wrap_parameters_options
-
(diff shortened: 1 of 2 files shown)


lib/rage/controller/parameters_wrappers/strong_parameters.rb
```diff
@@ -0,0 +1,11 @@
+require 'singleton'
+
+module ParametersWrappers
+  class StrongParameters
+    include Singleton
```
alex-rogachev, line 5: Added singleton here since there is no need to create multiple instances of this class for each controller because it doesn't store any state

lib/rage/router/backend.rb
```diff
@@ -100,6 +100,14 @@ def lookup(env)
     find(env, constraints)
   end
 
+  def set_params_wrappers(controller = RageController::API)
+    controller.__set_params_wrappers
+
+    controller.subclasses.each do |subclass|
+      set_params_wrappers(subclass)
+    end
+  end
```
alex-rogachev, line 109: This part recursively initiates parameters wrapping stack for all controllers starting with RageController::API and down by the Controller subclasses tree

lib/rage/setup.rb
```diff
@@ -8,6 +8,8 @@
 # Load application classes
 Rage.code_loader.setup
 
+Rage.__router.set_params_wrappers
```
alex-rogachev, line 11: Initiate parameters wrappers setting after controller classes are loaded and flags that control wrappers plugging in are set.

spec/controller/api/wrap_parameters_spec.rb
```diff
@@ -0,0 +1,319 @@
...
+        it 'wraps the param that is set to be included' do
+          initial_params = {param_a: :value, param_b: :value}
+          expected_result = {param_a: :value, param_b: :value, root: {param_a: :value}}
+
+          response = run_action(controller, :index, params: initial_params, env: {'CONTENT_TYPE' => "application/json"})
+          expect(response).to match([200, instance_of(Hash), [expected_result.to_json]])
+        end
+      end
+
+      context 'and :include option is set as Array' do
```
rsamoilov, line 92: Could you please also add a test case for when the `:include` option is not present in the params?

spec/controller/api/wrap_parameters_spec.rb
```diff
@@ -0,0 +1,319 @@
...
+        it 'wraps the params except the param that is set to be excluded' do
+          initial_params = {param_a: :value, param_b: :value}
+          expected_result = {param_a: :value, param_b: :value, root: {param_b: :value}}
+
+          response = run_action(controller, :index, params: initial_params, env: {'CONTENT_TYPE' => "application/json"})
+          expect(response).to match([200, instance_of(Hash), [expected_result.to_json]])
+        end
+      end
+
+      context 'and :exclude option is set as Array' do
```
rsamoilov, line 139: Could you please also add a test case for when the `:exclude` option is not present in the params?

rsamoilov (commented):
Great job, Alex!! Looking forward to merging this 🥳

lib/rage/controller/api.rb
```diff
@@ -77,6 +77,25 @@ def __register_action(action)
...
       activerecord_loaded = defined?(::ActiveRecord)
 
+      wrap_parameters_chunk = if __wrap_parameters_key
+        <<~RUBY
+          wrap_key = self.class.__wrap_parameters_key
+          if !@__params.key?(wrap_key) && @__env['CONTENT_TYPE']
+            wrap_options = self.class.__wrap_parameters_options
+  
+            wrapped_params = if wrap_options[:include]
+              @__params.slice(*[wrap_options[:include]].flatten)
```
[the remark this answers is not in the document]
alex-rogachev: yeah, I added flatten to support both symbol and array options, like `exclude: :param` and `exclude: [:param1, :param2]` but now I think that this is redundant and Rails supports only array options anyway so I'm going to remove flatten

lib/rage/controller/api.rb
```diff
@@ -77,6 +77,25 @@ def __register_action(action)
...
+            elsif wrap_options[:exclude]
+              @__params.except(*[wrap_options[:exclude]].flatten)
+            else
+              @__params
+            end
+  
+            @__params = @__params.merge({wrap_key => wrapped_params})
```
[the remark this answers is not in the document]
alex-rogachev: totally makes sense 👍

lib/rage/controller/api.rb
```diff
@@ -277,6 +300,20 @@ def skip_before_action(action_name, only: nil, except: nil)
...
+    # Initialize controller params wrapping into a nested hash.
+    # If initialized, params wrapping logic will be added to the controller.
+    #
+    # @param key [Symbol] key that the params hash will nested under
+    # @param options [Hash] wrapping options
+    # @example
+    #   wrap_parameters :user, include: [:name, :age]
+    # @example
+    #   wrap_parameters :user, exclude: :address
+    def wrap_parameters(key, options = {})
```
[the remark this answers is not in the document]
alex-rogachev: No, to be honest I thought using them but then just looked at how this is implemented in Rails and decided to leave it. I see that keyword arguments are more self-explanatory so I will apply them.

PR #97, opened 2024-08-10
Add request methods to the Rage::Request class.

cuneyter:
### Description
This PR adds four new methods to the `Rage::Request` class to enhance its functionality. The methods added are `url`, `path`, `fullpath`, and `user_agent`.

### Issue Addressed
Issue: Add request methods
The `Rage::Request` class needs additional methods to provide more detailed information about the request. These methods include:
- `url`: Returns the full URL of the request.
- `path`: Returns the path of the request.
- `fullpath`: Returns the full path including the query string.
- `user_agent`: Returns the user agent of the request by looking at the `HTTP_USER_AGENT` header.

### Changes Made
1. **`url` Method**:
   - Returns the full URL of the request.
   - Example: `request.url => "http://localhost:3000/users?show_archived=true"`

2. **`path` Method**:
   - Returns the path of the request.
   - Example: `request.path => "/users"`

3. **`fullpath` Method**:
   - Returns the full path including the query string.
   - Example: `request.fullpath => "/users?show_archived=true"`

4. **`user_agent` Method**:
   - Returns the user agent of the request by looking at the `HTTP_USER_AGENT` header.
   - Example: `request.user_agent => "Mozilla/5.0 (Macintosh; ..."`

5. **RSpec Test**:
   - Added tests to check if the methods are returning the expected result. 

### Documentation

Examples for each method have been added to the Rage::Request class.

### DIFF
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -37,6 +37,45 @@ class Rage::Request
     )
   end
 
+  # Returns the full URL of the request.
+  # @example
+  #   request.url # => "https://example.com/users?show_archived=true"
+  def url
+    scheme = @env["rack.url_scheme"]
+    host = @env["SERVER_NAME"]
+    port = @env["SERVER_PORT"]
+    path = @env["PATH_INFO"]
+    query_string = @env["QUERY_STRING"]
+
+    port_part = (scheme == "http" && port == "80") || (scheme == "https" && port == "443") ? "" : ":#{port}"
+    query_part = query_string.empty? ? "" : "?#{query_string}"
+
+    "#{scheme}://#{host}#{port_part}#{path}#{query_part}"
+  end
+
+  # Returns the path of the request.
+  # @example
+  #   request.path # => "/users"
+  def path
+    @env["PATH_INFO"]
+  end
+
+  # Returns the full path including the query string.
+  # @example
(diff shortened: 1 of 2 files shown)


lib/rage/request.rb
```diff
@@ -37,6 +37,45 @@ def fresh?(etag:, last_modified:)
...
+    path = @env["PATH_INFO"]
+    query_string = @env["QUERY_STRING"]
+    query_string.empty? ? path : "#{path}?#{query_string}"
+  end
+
+  # Returns the user agent of the request.
+  # @example
+  #  request.user_agent # => "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36"
+  def user_agent
+    headers["HTTP_USER_AGENT"]
```
rsamoilov, line 76:
```suggestion
    @env["HTTP_USER_AGENT"]
```

How about we use `@env` here? This is exactly what `headers` will do anyway.

cuneyter: @rsamoilov - thank you for your recommendation. I have updated it to @env

rsamoilov (approved):
👍

PR #99, opened 2024-08-18
Implement `middleware` and `version` commands in the rage/cli.rb file.

cuneyter:
### Description
This PR updates Rage’s CLI utility to add the `rage middleware` and `rage version` commands to enhance its functionality.

### Issue Addressed
**Issue**: Add middleware and version commands
The Rage::CLI class needs additional commands to provide more detailed information about the application. These commands include:

- `middleware`: Lists the Rack middleware stack enabled for the application.
- `version`: Returns the current version of the framework.

### Changes Made
`middleware` Command:

Lists the Rack middleware stack enabled for the application.
Example: `rage middleware` => "Rage::FiberWrapper Rage::Reloader"

`version` Command:

Returns the current version of the framework.
Example: `rage version` => "1.0.0"

### RSpec Test:

Added tests to check if the commands are returning the expected result.

### Documentation
Examples for each command have been added to the Rage::CLI class.

### DIFF
--- a/lib/rage/cli.rb
+++ b/lib/rage/cli.rb
@@ -2,6 +2,7 @@
 
 require "thor"
 require "rack"
+require "rage/version"
 
 module Rage
   class CLI < Thor
@@ -110,6 +111,29 @@ module Rage
       environment
       ARGV.clear
       IRB.start
+    end
+
+    desc "middleware", "List Rack middleware stack enabled for the application"
+    def middleware
+      app = ::Rack::Builder.parse_file(options[:config] || "config.ru")
+      app = app[0] if app.is_a?(Array)
+      middlewares = []
+
+      while app
+        if app.instance_variable_defined?(:@app)
+          middlewares << app.class
+          app = app.instance_variable_get(:@app)
+        else
+          break
+        end
+      end
+
+      middlewares.each { |middleware| puts middleware }
+    end
+
+    desc "version", "Return the current version of the framework"
+    def version
+      puts Rage::VERSION
     end
 
     private
--- a/lib/rage/cli.rb
+++ b/lib/rage/cli.rb
@@ -115,20 +115,11 @@ module Rage
 
     desc "middleware", "List Rack middleware stack enabled for the application"
     def middleware
-      app = ::Rack::Builder.parse_file(options[:config] || "config.ru")
-      app = app[0] if app.is_a?(Array)
-      middlewares = []
+      environment
 
-      while app
-        if app.instance_variable_defined?(:@app)
-          middlewares << app.class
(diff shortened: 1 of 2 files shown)


lib/rage/cli.rb
```diff
@@ -112,6 +113,29 @@ def console
       IRB.start
     end
 
+    desc "middleware", "List Rack middleware stack enabled for the application"
+    def middleware
+      app = ::Rack::Builder.parse_file(options[:config] || "config.ru")
+      app = app[0] if app.is_a?(Array)
+      middlewares = []
+
+      while app
```
rsamoilov, line 122: You might want to use `Rage.config.middleware.middlewares` instead.

lib/rage/cli.rb
```diff
@@ -112,6 +113,29 @@ def console
...
+        if app.instance_variable_defined?(:@app)
+          middlewares << app.class
+          app = app.instance_variable_get(:@app)
+        else
+          break
+        end
+      end
+
+      middlewares.each { |middleware| puts middleware }
```
rsamoilov, line 131: Can we make it look a bit closer to the output of `rails middleware`?

lib/rage/cli.rb
cuneyter: @rsamoilov - Thank you for your comment. 
What do you think about the below approach? 
```ruby
def middleware
  environment

  Rage.config.middleware.middlewares.each do |middleware|
    say "use #{middleware.first.inspect}"
  end
end
```
This method gives me the following output in the rage test app 
```
➜  rage-test-app git:(master) ✗ rage middleware
use Rage::FiberWrapper
use Rage::Reloader
use Rage::Cors
```
Is it ok to call environment here or is there any better way to initialise the Rage::Configuration so that `Rage.config` is available in the middleware method?

lib/rage/cli.rb
cuneyter: @rsamoilov - Thank you for the comment. I think the line will produce a similar output. What do you think?
```
say "use #{middleware.first.inspect}"
```

lib/rage/cli.rb
rsamoilov: Yeah, calling `environment` should be ok

lib/rage/cli.rb
rsamoilov: Looks good. Probably calling `name` would be a bit more clear.

```
say "use #{middleware.first.name}"
```

lib/rage/cli.rb
cuneyter: Great. Thank you.

rsamoilov (approved):
👍

PR #100, opened 2024-08-18
Static file server

rsamoilov:
### Description

This adds the ability for Rage apps to serve static assets from the `public` folder.

### Usage

Enable the feature using the `public_file_server` configuration:

```ruby
  Rage.configure do
    config.public_file_server.enabled = true
  end
```

### Features

* Support [Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) requests;
* Automatically serve `gzip` version of files if the client supports the `gzip` transfer-encoding;
* Support [If-None-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) and [If-Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range) headers;

### Benchmark

`wrk http://localhost:3000/assets/40kb.js.gz`

**Rails:** 8719 RPS
**Rage:** 22980 RPS


rsamoilov:
@nikki-dev-s This will go into the 1.9.0 release.

PR #123, opened 2025-01-16
Enhance Request

aaoafk:
hello,

i think this should mostly work unless i've missed something.

the only issue right now is maybe `remote_ip` needs to be thought through a little more and the current method of checking the http action does not verify against `HTTP_METHODS` I just need to port that underscore method over or just do something else.

lmk your initial thoughts

### DIFF
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -5,28 +5,6 @@ require "rack/request"
 require "forwardable"
 
 class Rage::Request
-  extend Forwardable
-  include Rack::Request::Helpers
-  include Rack::Request::Env
-  attr_reader :rack_request
-  def_delegators :@rack_request,
-                 :ssl?,
-                 :host,
-                 :port,
-                 :query_string,
-                 :env,
-                 :get_header,
-                 :get?,
-                 :post?,
-                 :patch?,
-                 :put?,
-                 :delete?,
-                 :head?,
-                 :url
-
-  def_delegator :@rack_request, :content_type, :format
-  def_delegator :@rack_request, :ip, :remote_ip
-
   IP_HOST_REGEXP  = /\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/
   # HTTP methods from [RFC 2616: Hypertext Transfer Protocol -- HTTP/1.1](https://www.ietf.org/rfc/rfc2616.txt)
   RFC2616 = %w(OPTIONS GET HEAD POST PUT DELETE TRACE CONNECT)
@@ -45,15 +23,16 @@ class Rage::Request
   # HTTP methods from [RFC 5789: PATCH Method for HTTP](https://www.ietf.org/rfc/rfc5789.txt)
   RFC5789 = %w(PATCH)
 
-  HTTP_METHODS = (RFC2616 + RFC2518 + RFC3253 + RFC3648 + RFC3744 + RFC5323 + RFC4791 + RFC5789).to_set
+  HTTP_METHODS_SET = (RFC2616 + RFC2518 + RFC3253 + RFC3648 + RFC3744 + RFC5323 + RFC4791 + RFC5789).to_set
 
+  attr_accessor :env
   # @private
   def initialize(env, custom_proxies: nil)
-    super(env)
-    @rack_request = Rack::Request.new(env)
+    @env = env
+    # super(env)
     after_initialize(custom_proxies) if custom_proxies
   end
-  
+
   def after_initialize(custom_proxies)
     if custom_proxies&.is_a?(Regexp)
       Rack::Request.class_exec do |rage_trusted_proxies|
@@ -61,8 +40,83 @@ class Rage::Request
       end
     else
       raise(Rage::Errors::InvalidCustomProxy, "Custom proxy should be a regexp. You passed in a #{custom_proxies.class}")
-
     end
+  end
+
+  def rack_request
+    @rack_request ||= Rack::Request.new(@env)
+  end
+
+  def ssl?
+    rack_request.ssl?
+  end
+
+  def host
+    rack_request.host
+  end
+
+  def port
+    rack_request.port
+  end
+
+  def query_string
+    rack_request.query_string
+  end
+
+  def env
+    rack_request.env
+  end
+
+  def get_header(name)
+    rack_request.get_header(name)
+  end
+
+  def get?
+    rack_request.get?
+  end
+
+  def post?
+    rack_request.post?
+  end
+
+  def patch?
+    rack_request.patch?
+  end
+
+  def put?
+    rack_request.put?
+  end
+
+  def delete?
+    rack_request.delete?
+  end
+
+  def head?
+    rack_request.head?
+  end
+
+  def url
+    rack_request.url
+  end
+
+  def path
+    rack_request.path
+  end
+
+  def fullpath
+    rack_request.fullpath
+  end
+
+  def user_agent
+    rack_request.user_agent
+  end
+
+  def format
+    rack_request.content_type
+  end
+
+  def remote_ip
+    rack_request.ip
   end
 
   # Get the request headers.
@@ -97,25 +151,25 @@ class Rage::Request
   # Returns the path of the request.
   # @example
   #   request.path # => "/users"
-  def path
-    @env["PATH_INFO"]
-  end
+  # def path
+  #   @env["PATH_INFO"]
+  # end
 
   # Returns the full path including the query string.
   # @example
   #   request.fullpath # => "/users?show_archived=true"
-  def fullpath
-    path = @env["PATH_INFO"]
-    query_string = @env["QUERY_STRING"]
-    query_string.empty? ? path : "#{path}?#{query_string}"
-  end
+  # def fullpath
+  #   path = @env["PATH_INFO"]
+  #   query_string = @env["QUERY_STRING"]
+  #   query_string.empty? ? path : "#{path}?#{query_string}"
+  # end
 
   # Returns the user agent of the request.
   # @example
   #  request.user_agent # => "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36"
-  def user_agent
-    @env["HTTP_USER_AGENT"]
-  end
+  # def user_agent
+  #   @env["HTTP_USER_AGENT"]
+  # end
 
   def domain(tld_length = 1)
     extract_domain(host, tld_length)
@@ -138,12 +192,12 @@ class Rage::Request
   private
 
   def check_method(name)
-    http_methods_set = HTTP_METHODS
+    http_methods_set = HTTP_METHODS_SET
     if name
       if http_methods_set.include?(name)
         name
       else
-        raise(Rage::Errors::UnknownHTTPMethod, "#{name}, accepted HTTP methods are #{http_methods_array}")
+        raise(Rage::Errors::UnknownHTTPMethod, "#{name}, accepted HTTP methods are #{http_methods_set.to_a}")
       end
     end
   end
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -2,17 +2,17 @@
 
 require "time"
 require "rack/request"
-require 'forwardable'
+require "forwardable"
 
 class Rage::Request
   extend Forwardable
-  attr_accessor :rack_request
-  # get?/post?/patch?/put?/delete?/head?
+  include Rack::Request::Helpers
+  include Rack::Request::Env
+  attr_reader :rack_request
   def_delegators :@rack_request,
                  :ssl?,
                  :host,
                  :port,
-                 :query,
                  :query_string,
                  :env,
                  :get_header,
@@ -21,7 +21,11 @@ class Rage::Request
                  :patch?,
                  :put?,
                  :delete?,
-                 :head?
+                 :head?,
+                 :url
+
+  def_delegator :@rack_request, :content_type, :format
+  def_delegator :@rack_request, :ip, :remote_ip
 
   IP_HOST_REGEXP  = /\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/
   # HTTP methods from [RFC 2616: Hypertext Transfer Protocol -- HTTP/1.1](https://www.ietf.org/rfc/rfc2616.txt)
@@ -41,20 +45,24 @@ class Rage::Request
   # HTTP methods from [RFC 5789: PATCH Method for HTTP](https://www.ietf.org/rfc/rfc5789.txt)
   RFC5789 = %w(PATCH)
 
-  HTTP_METHODS = RFC2616 + RFC2518 + RFC3253 + RFC3648 + RFC3744 + RFC5323 + RFC4791 + RFC5789
-
-  HTTP_METHOD_LOOKUP = {}
-
-  # Populate the HTTP method lookup cache.
-  # TODO: Extract underscore method
-  # HTTP_METHODS.each { |method|
-  #   HTTP_METHOD_LOOKUP[method] = method.underscore.to_sym
-  # }
+  HTTP_METHODS = (RFC2616 + RFC2518 + RFC3253 + RFC3648 + RFC3744 + RFC5323 + RFC4791 + RFC5789).to_set
 
   # @private
-  def initialize(env)
-    @env = env
+  def initialize(env, custom_proxies: nil)
+    super(env)
     @rack_request = Rack::Request.new(env)
+    after_initialize(custom_proxies) if custom_proxies
+  end
+  
+  def after_initialize(custom_proxies)
+    if custom_proxies&.is_a?(Regexp)
+      Rack::Request.class_exec do |rage_trusted_proxies|
+        # hook on trusted_proxy? 
+      end
+    else
+      raise(Rage::Errors::InvalidCustomProxy, "Custom proxy should be a regexp. You passed in a #{custom_proxies.class}")
+
+    end
   end
 
   # Get the request headers.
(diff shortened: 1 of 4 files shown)


lib/rage/request.rb
```diff
@@ -76,8 +125,54 @@ def user_agent
...
+      # )
+      @method ||= get_header("rack.methodoverride.original_method") || get_header("REQUEST_METHOD")
+    else
+      super
+    end
+  end
+
+  def remote_ip(*args)
+    @env['HTTP_X_FORWARDED_FOR']
+  end
```
rsamoilov, line 152: I agree this is going to be more complicated. I would encourage you to work on `remote_ip` in a separate PR, but that's up to you.

Essentially, we want the algorithm to be close to [what it is in Rails](https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html) with the ability to customize the list of trusted proxies.

rsamoilov (commented):
I've left some comments, but this looks great!

lib/rage/request.rb
```diff
@@ -1,11 +1,60 @@
...
 
 class Rage::Request
+  extend Forwardable
+  attr_accessor :rack_request
+  # get?/post?/patch?/put?/delete?/head?
+  def_delegators :@rack_request,
+                 :ssl?,
+                 :host,
+                 :port,
+                 :query,
```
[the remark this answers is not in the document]
aaoafk: I removed `:query` dunno where I got that from lol

aaoafk:
Also one thing I added to cover `format` as defined in rails api is just:

```
  def_delegator :@rack_request, :content_type, :format
```

aaoafk:
As an aside. I noticed that we have `url` implemented inside: `Rage::Request` like so:

```
  # Returns the full URL of the request.
  # @example
  #   request.url # => "https://example.com/users?show_archived=true"
  def url
    scheme = @env["rack.url_scheme"]
    host = @env["SERVER_NAME"]
    port = @env["SERVER_PORT"]
    path = @env["PATH_INFO"]
    query_string = @env["QUERY_STRING"]

    port_part = (scheme == "http" && port == "80") || (scheme == "https" && port == "443") ? "" : ":#{port}"
    query_part = query_string.empty? ? "" : "?#{query_string}"

    "#{scheme}://#{host}#{port_part}#{path}#{query_part}"
  end
```

Is there a reason we don't want to use `url` behavior defined inside rack?

```
      # Tries to return a remake of the original request URL as a string.
      def url
        base_url + fullpath
      end
```

lib/rage/request.rb
```diff
@@ -1,11 +1,60 @@
...
+  # Populate the HTTP method lookup cache.
+  # TODO: Extract underscore method
+  # HTTP_METHODS.each { |method|
+  #   HTTP_METHOD_LOOKUP[method] = method.underscore.to_sym
+  # }
+
   # @private
   def initialize(env)
     @env = env
+    @rack_request = Rack::Request.new(env)
```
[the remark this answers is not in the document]
rsamoilov: I think I like your approach a bit more because it allows us to add to the class only the methods we are interested in.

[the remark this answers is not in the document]
rsamoilov: Is it possible to create `@rack_request` lazily?

lib/rage/request.rb
```diff
@@ -76,8 +125,54 @@ def user_agent
...
+  def method(*args)
+    if args.empty?
+      # @method ||= check_method(
+      #   get_header("rack.methodoverride.original_method") || get_header("REQUEST_METHOD")
+      # )
+      @method ||= get_header("rack.methodoverride.original_method") || get_header("REQUEST_METHOD")
+    else
+      super
+    end
+  end
```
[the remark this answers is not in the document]
rsamoilov: Got it, makes sense. Should `HTTP_METHODS` be a `Set` then for faster lookups?

lib/rage/request.rb
rsamoilov: Both options work. I guess modifying `self.ip_filter` would be somewhat more expected.

lib/rage/request.rb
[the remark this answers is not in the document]
aaoafk: I don't think so. I tried it out and it looks like there might be some weird behavior with lazy loading and def_delegator, it was `nil` for all my specs, even with a:

```
def rack_request
  @rack_request ||= Rack::Request.new(env)
end
```

lib/rage/request.rb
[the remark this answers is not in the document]
aaoafk: I changed this to a set. I benchmarked it and it looks like it was 2x as fast for lookup

lib/rage/request.rb
aaoafk: I modified the `initialize` method for `request` is that what you had in mind?

lib/rage/request.rb
```diff
@@ -76,8 +117,50 @@ def user_agent
...
+
+  def protocol
+    (ssl? == "https://") ? "https://" : "http://"
+  end
+
+  def method(*args)
+    if args.empty?
+      @method ||= check_method(
+        get_header("rack.methodoverride.original_method") || get_header("REQUEST_METHOD")
```
rsamoilov, line 131: Why do we need to support `rack.methodoverride.original_method` here? I am trying to send a request to a Rage app with the `X-HTTP-Method-Override` header, but `rack.methodoverride.original_method` is still not getting set.

lib/rage/request.rb
```diff
@@ -2,17 +2,17 @@
...
 require "rack/request"
-require 'forwardable'
+require "forwardable"
 
 class Rage::Request
   extend Forwardable
-  attr_accessor :rack_request
-  # get?/post?/patch?/put?/delete?/head?
+  include Rack::Request::Helpers
+  include Rack::Request::Env
```
rsamoilov, line 10: So, if you include these two modules, does it mean we don't need the `def_delegators` calls and the `@rack_request` variable anymore?

In any case, we should explicitly document the new methods, as YARD doesn't pick up neither the `def_delegators` calls nor the methods from the included modules (you can verify by running `yard server --reload` and then visiting `http://localhost:8808/docs/Rage/Request`)

lib/rage/request.rb
rsamoilov: Pretty much, but it should be part of the [configuration](https://github.com/rage-rb/rage/blob/master/lib/rage/configuration.rb) instead, same way it is in Rails - https://guides.rubyonrails.org/configuring.html#actiondispatch-remoteip.

rsamoilov:
Hey @aaoafk , really great job so far! Sorry I wasn't able to provide prompt feedback, but I promise I will get better 😄

lib/rage/request.rb
aaoafk: I'll look into what I imagine to be that little configuration object but if memory serves correct I'm not sure that this module is being used anymore ill take another look, because that module resolves the remote ip but ActionDispatch request is forwarding it to rack

aaoafk:
> Hey @aaoafk , really great job so far! Sorry I wasn't able to provide prompt feedback, but I promise I will get better 😄

that's fine im more so just trying to get better at programming so your reviews are helping no pressure im working on this in my free time lol

lib/rage/request.rb
aaoafk: Virtually yes, we wouldn't need it. I think I would need to bring some methods down in that case if we got rid of `@rack_request` tho, would need to double check

lib/rage/request.rb
aaoafk: I'll need to look into `rack.methodoverride.original_method` a little more closely... I'm guessing that header you set is supposed to lead to a different result, pretty much just ripped that right from rails didn't think too much about it.

rsamoilov:
Hey @aaoafk , how's it going? Is there anything I can help you with?

aaoafk:
Hey @rsamoilov recently had a big life change, moved to Puerto Rico for work so I was pretty busy the last couple weeks sorting everything out. I'll try to take a look at this soon.

lib/rage/request.rb
aaoafk: So to close this out we have two options:

1) `@rack_request` with `def_delegator`.

2) include Rack::Request::Helpers and include Rack::Request::Env approach. This also works the same way, the same methods are covered but we will need to bring some methods down to comply with rails api. Look at this test output with composing the modules behavior:

```
Rage::Request
  returns the full URL
  returns the path
  returns the full path with query string
  returns the user agent
  handles missing user agent header
  handles the host property of a request
  handles the domain property of a request
  handles the method property of a request
  handles `get?` HTTP verb of a request
  handles `post?` HTTP verb of a request
  handles the `patch?` HTTP verb of a request
  handles the `put?` HTTP verb of a request
  handles the `delete?` HTTP verb of a request
  handles the `head?` HTTP verb of a request
  handles the port property of a request
  handles the protocol property of a request
  handles the query string property of a request
  handles the remote ip property of a request (FAILED - 1)
  handles the env property of a request
  handles the format property of a request (FAILED - 2)
```
Remote ip & format need to be brought down.

Having both `@rack_request` and the inclusion of these modules doesn't really make sense to me because it seems to be effectively duplicating behavior.

aaoafk: > So, if you include these two modules, does it mean we don't need the `def_delegators` calls and the `@rack_request` variable anymore?
> 
> In any case, we should explicitly document the new methods, as YARD doesn't pick up neither the `def_delegators` calls nor the methods from the included modules (you can verify by running `yard server --reload` and then visiting `http://localhost:8808/docs/Rage/Request`)

Based on the approach we take I will make sure that the documentation is `YARD` compatible.

rsamoilov: What about using `@rack_request` **without** `def_delegators`?


It would be a bit more verbose, but on the other hand, we’d have all methods we are interested in defined explicitly, which means less hassle with YARD. And we would also be able to create the `@rack_request` object lazily.

aaoafk: Do you mean something like:

```
def host
  @rack_request.host
end
```

If so we would at least need to change the `env` inside `Rage::Request` to be an instance variable after which we can do: 

```
def rack_request
  @rack_request ||= Rack::Request.new(@env)
end
```

Since it needs `env`

rsamoilov: Yeah, that's what I meant. WDYT?

aaoafk: Just did the refactor and that works fine. Let's go with this then.

aaoafk:
I committed changes for you to view I still need to look at `rack.methodoverride.original_method` and the `custom_proxies` stuff

lib/rage/request.rb
aaoafk: So I was able to get this to work if I use the `Rack::MethodOverride` module inside a `Rage.configure` block:

```
Rage.configure do
  # Specify the number of server processes to run. Defaults to number of CPU cores.
  config.server.workers_count = 1

  # Specify the port the server will listen on.
  config.server.port = 3000

  # Specify the logger
  config.logger = Rage::Logger.new(STDOUT)

  config.middleware.use Rage::Reloader
  config.middleware.use Rack::MethodOverride
end
```

lib/rage/request.rb
```diff
@@ -1,11 +1,122 @@
...
+  def port
+    rack_request.port
+  end
+
+  def query_string
+    rack_request.query_string
+  end
+
+  def env
+    rack_request.env
```
rsamoilov, line 67:
```suggestion
    @env
```

lib/rage/request.rb
```diff
@@ -1,11 +1,122 @@
...
+        # hook on trusted_proxy? 
+      end
+    else
+      raise(Rage::Errors::InvalidCustomProxy, "Custom proxy should be a regexp. You passed in a #{custom_proxies.class}")
+    end
+  end
+
+  def rack_request
+    @rack_request ||= Rack::Request.new(@env)
+  end
```
rsamoilov, line 48: Could we make this method private?

lib/rage/request.rb
```diff
@@ -1,11 +1,122 @@
 # frozen_string_literal: true
 
 require "time"
+require "rack/request"
+require "forwardable"
```
rsamoilov, line 5:
```suggestion
```

lib/rage/request.rb
```diff
@@ -1,11 +1,122 @@
 # frozen_string_literal: true
 
 require "time"
+require "rack/request"
```
rsamoilov, line 4:
```suggestion
```

Do we need to manually require `rack/request`?

lib/rage/request.rb
rsamoilov: I still don't think we need it. What's the intended functionality here?

lib/rage/request.rb
```diff
@@ -121,7 +258,7 @@ def [](requested_header)
         if "CONTENT_TYPE" == requested_header || "CONTENT_LENGTH" == requested_header
           @env[requested_header]
         else
-          @env["#{HTTP}#{requested_header}"]
+          @env["#{HTTP_}#{requested_header}"]
```
rsamoilov, line 261:
```suggestion
          @env["#{HTTP}#{requested_header}"]
```

rsamoilov:
> I committed changes for you to view I still need to look at `rack.methodoverride.original_method` and the `custom_proxies` stuff

It lacks YARD docs, but apart from that, this bit looks good 👍

lib/rage/request.rb
aaoafk: AFAIK it is to get the correct HTTP method if its overwritten: `curl -v -X POST http://localhost:3000/ -H "X-HTTP-Method-Override: GET"`

If `X-HTTP-Method-Override` is set then get the HTTP verb that it sets there. Otherwise, grab the request method which is `POST`.

If a client wants to use this behavior then they would have to use `Rack::MethodOverride` middleware. Otherwise, we always get the request method, even if its overwritten by that header... Does that make sense?

lib/rage/request.rb
aaoafk: You're right we don't need that require rack is already a dependency its redundant.

lib/rage/request.rb
aaoafk: yes we can and probably should I moved it to private section

lib/rage/request.rb
rsamoilov: Aha, it makes sense now!

aaoafk:
@rsamoilov are you imagining that the `custom_proxies` bit is a part of `config.server.custom_proxies` or `config.custom_proxies` I think the former seems properly scoped

rsamoilov:
> @rsamoilov are you imagining that the `custom_proxies` bit is a part of `config.server.custom_proxies` or `config.custom_proxies` I think the former seems properly scoped

`config.server` is more about the server process - we can use it to set the port the server is running on or the number of workers. How about we add a new `config.http` scope and put `custom_proxies` (should actually be `trusted_proxies`) in there?

aaoafk:
Yeah I was thinking it might take a little bit for that one I'll remove the remote ip stuff and do a push tonight

aaoafk:
Hey Roman. I think this should have everything around the remote_ip removed and I corrected the merge conflict, I think we just wanted to take changes from both files since they weren't really conflicting.

lib/rage/request.rb
```diff
@@ -3,11 +3,143 @@
...
+  RFC3744 = %w(ACL)
+  # HTTP methods from [RFC 5323: WebDAV SEARCH](https://www.ietf.org/rfc/rfc5323.txt)
+  RFC5323 = %w(SEARCH)
+  # HTTP methods from [RFC 4791: Calendaring Extensions to WebDAV](https://www.ietf.org/rfc/rfc4791.txt)
+  RFC4791 = %w(MKCALENDAR)
+  # HTTP methods from [RFC 5789: PATCH Method for HTTP](https://www.ietf.org/rfc/rfc5789.txt)
+  RFC5789 = %w(PATCH)
+
+  # Set data structure of all RFC defined HTTP headers
+  HTTP_METHODS_SET = (RFC2616 + RFC2518 + RFC3253 + RFC3648 + RFC3744 + RFC5323 + RFC4791 + RFC5789).to_set
```
rsamoilov, line 26: It looks like this line fails on Ruby 3.1. You can fix it by adding the following line to the top of the file:

```
require "set"
```

spec/rage/request_spec.rb
```diff
@@ -63,7 +65,63 @@
...
+    expect(request.head?).not_to be_nil 
+  end
+
+  it "handles the port property of a request" do
+    expect(request.port).not_to be_nil
+  end
+
+  it "handles the protocol property of a request" do
+    expect(request.protocol).not_to be_nil
+  end
```
rsamoilov, line 111: Did you remove the `protocol` method from the `Request` class on purpose? This test is currently failing because there's no such method.

rsamoilov (commented):
Hey Steven, I've left several comments to fix the tests.

Could you also fix the [style](https://github.com/rage-rb/rage/actions/runs/13823976262/job/38732136135?pr=123) warnings? Once it's done, the PR will be good to go.

spec/rage/request_spec.rb
aaoafk: Strange. I added it back. I don't remember deleting it.

aaoafk:
One test is failing now and I think it might be due to a stale value for the `SERVER_PORT` inside this test:

```
  it "handles default ports in URL" do
    env["SERVER_PORT"] = "80"
    expect(request.url).to eq("http://localhost/users?show_archived=true")

    env["rack.url_scheme"] = "https"
    env["SERVER_PORT"] = "443"
    expect(request.url).to eq("https://localhost/users?show_archived=true")
  end
```

The output is:

expected: "http://localhost/users?show_archived=true"
got: "http://localhost:3000/users?show_archived=true"

I think this is because of the memoized request, it is using an old value for env, since the test does a direct modification on env it wouldn't be reflected there, but some rack methods do query headers directly from env so...

rsamoilov:
Aha, good catch! Those tests should use contexts instead. Do you want me to fix those?

aaoafk:
Do you mean rspec contexts ?

rsamoilov:
Yep.

aaoafk:
How does this look?

rsamoilov:
Rubocop is failing, but it looks like it's picked up the wrong commit.

aaoafk:
🔥 I'll see about remote ip next or something else curious about the implementation

Issue #115, opened 2025-01-09
🐞 Using `Fiber.schedule` in console


rsamoilov:
### Description

Currently, because Rage scheduler is only available [during the request lifecycle](https://github.com/rage-rb/rage/blob/master/lib/rage/middleware/fiber_wrapper.rb), calling `Fiber.schedule` in the console raises the `No scheduler is available! (RuntimeError)` error.

### Steps to Reproduce

- Create a new Rage application.
- Open Rage console using the `rage c` command.
- Run `Fiber.schedule { sleep(1); puts("hello") }`.
- Actual behaviour: `No scheduler is available! (RuntimeError)` is raised.
- Expected behaviour: the fiber is executed in the synchronous manner.

### Tips

- See [lib/rage/rspec.rb](https://github.com/rage-rb/rage/blob/master/lib/rage/rspec.rb) for a reference implementation.
- Check the [overview doc](https://github.com/rage-rb/rage/blob/master/OVERVIEW.md) that shows how Rage's core components interact with each other.
- **Feel free to ask any questions or request help in the comments below!**

TheBlackArroVV:
Hey, @rsamoilov, please, take a look into https://github.com/rage-rb/rage/pull/125 when you have time

PR #125, opened 2025-01-26
Add default FiberScheduler to rage console

TheBlackArroVV:
Closes https://github.com/rage-rb/rage/issues/115


rsamoilov (commented):
Hey @TheBlackArroVV , thank you for the PR! This is pretty much what we are looking for, but it doesn't work correctly in all cases.

It works fine with synchronous fibers (e.g. `Fiber.schedule { rand }`) but doesn't work correctly with asynchronous.

For example, running the following code in the console raises `attempt to yield on a not resumed fiber (FiberError)`:

```ruby
require "net/http"
Fiber.await(Fiber.schedule { Net::HTTP.get(URI("http://httpbin.org/delay/1")) })
```

The reason for that is that `Rage::FiberScheduler` is using the server's reactor - if the server is not running, the scheduler will not work. Thus, we will likely need a mock scheduler here, similar to the one we use in [tests](https://github.com/rage-rb/rage/blob/master/lib/rage/rspec.rb).

TheBlackArroVV:
Hey, @rsamoilov,
I have updated it, should be good now. I tried to make it work with refinements, but, unfortunately looks like new console is outside of it's scope

lib/rage/cli.rb
```diff
@@ -3,6 +3,7 @@
 require "thor"
 require "rack"
 require "rage/version"
+require "rage/sync_fiber"
```
rsamoilov, line 6: Requiring this file here means it would also be applied when running the server. Instead, we would only want it to be applied when running the console, i.e. in the `console` method.

spec/rage/sync_fiber_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rage/sync_fiber"
```
rsamoilov, line 1: Same issue here - such way this file will be applied to all tests, even the ones where we test the actual scheduler, e.g. [spec/fiber_spec.rb](https://github.com/rage-rb/rage/blob/master/spec/fiber_spec.rb).

lib/rage/cli.rb
TheBlackArroVV: Fair, updated

lib/rage/cli.rb
```diff
@@ -4,6 +4,23 @@
...
+    fiber.resume
+
+    fiber
+  end
+
+  def self.await(fibers)
+    Array(fibers).map(&:__get_result)
+  end
+end
+
```
rsamoilov, line 23: Unfortunately, this is the same as requiring the file 😞 
This patch also applies to the server process, effectively making all code inside the application synchronous.

You can verify this by creating a new Rage app and adding an action that calls `sleep 1`. Sending however many requests in parallel to this action should only take one second, but with these changes Rage will be processing the requests one by one.

Instead, we only want these changes applied to the [console](https://github.com/rage-rb/rage/blob/master/lib/rage/cli.rb#L147-L156) method.

rsamoilov:
Closing as inactive. Feel free to reopen if you're going to keep working on this!

PR #130, opened 2025-02-11
Update app template to include all app rake tasks

pjb3:
When I generate a new rage app, it makes a lib/tasks directory. If I put a .rake file in there, the tasks defined in that file aren't available when I run rake. Should they be?

If I add this line to the Rakefile in the root of the project:

```
Dir[Rage.root.join("lib/tasks/**/*.rake")].each { |file| load file }
```

Then the tasks are available. If that is what users are expected to do, would it make sense to have that in that file then?

This pull request adds that to the template, but if there is a different way app rake tasks are intended to be loaded, let me know

### DIFF
--- a/lib/rage/tasks.rb
+++ b/lib/rage/tasks.rb
@@ -7,6 +7,7 @@ class Rage::Tasks
   class << self
     def init
       load_db_tasks if defined?(StandaloneMigrations)
+      load_app_tasks
     end
 
     private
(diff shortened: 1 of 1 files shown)


rsamoilov:
Hey Paul, thanks for the PR!

> If I put a .rake file in there, the tasks defined in that file aren't available when I run rake. Should they be?

They should - that's definitely an overlook on my side.

I would suggest putting this code [in here](https://github.com/rage-rb/rage/blob/83a59f600d0c60bbf6c5fd48ef14258e9b894859/lib/rage/tasks.rb#L8-L10). This way, our `Rakefile` will remain nice and tidy, and it will also work for old applications.

pjb3:
@rsamoilov Good suggestion. I've updated the commit to have the loading happen in `Rage.load_tasks`. Let me know if there are any other changes you would like.

rsamoilov:
Hey @pjb3 , this looks great!

PR #154, opened 2025-05-03
Router resource helper

lkibbalam:
Implements

HTTP Verb | Path | Controller#Action | Description
POST | /photo | photo#create | Create a new photo
GET | /photo | photo#show | Display the photo
PATCH/PUT | /photo | photo#update | Update the photo
DELETE | /photo | photo#destroy | Delete the photo

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,4 +22,5 @@ group :test do
   gem "websocket-client-simple"
   gem "prism"
   gem "redis-client"
+  gem "pry"
 end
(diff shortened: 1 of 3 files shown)


rsamoilov:
Hey @lkibbalam 

> Do we want just ignore param option like param: :slug . I would rather raise an error or warning

Could you please elaborate? Why would we want to ignore the `param` option? We do support it in the [resources](https://rage-rb.pages.dev/Rage/Router/DSL/Handler#resources-instance_method) helper.

> You mentioned /new path/action in the issue's description but didn't mention  /edit. Could you clarify if we need these actions

No, we don't want `edit` to be generated either.

> Do you mind if I make some small changes to both resource and resources because now they have 80% same code

Go for it.

rsamoilov:
Ah, right. Let's just ignore it.

PR #163, opened 2025-07-28
[OpenAPI] Validator Task

alex-rogachev:
- Add `openapi:validate` Rake task that builds the OpenAPI spec and fails (non-zero exit) when any tag warnings were produced during the build.
- Collect warnings only for the duration of a build (`Rage::OpenAPI.__collect_warnings`) so docs serving does not retain a process-wide warning list; abort if the app is not booted to avoid a false pass on an empty router.


lib/rage/openapi/openapi.rb
```diff
@@ -200,9 +200,20 @@ def self.__resolve_resource(klass_str, namespace)
...
   end
 
+  # @private
+  # @return [Array<String>] the warnings logged inside the block
+  def self.__collect_warnings
+    @__warnings = []
+    yield
+    @__warnings
+  ensure
+    @__warnings = nil
```
alex-rogachev, line 214: Hello @rsamoilov 
What do you think about implementing the warnings accumulator as a wrapper around `Rage::OpenAPI.build`? I feel like this approach works well because the additional logic doesn't add overhead to the original `__log_warn` implementation and remains optional, which aligns with the idea that new features shouldn't affect performance for users who don't use them.

rsamoilov: I like the current approach, but I'm not entirely sure what a wrapper around `Rage::OpenAPI.build` would look like.

I've approved the PR and will be happy to merge it. Let me know if you'd like to rebuild it though.

rsamoilov (approved):

rsamoilov:
FIY the approach is now a bit different, so I moved the validation Rake task under Rage::CLI - #382.

PR #169, opened 2025-11-05
Rage::Telemetry

rsamoilov:
`Rage::Telemetry` is a standartized interface that allows to instrument different parts of the framework. 

```ruby
class MyInstrumenter < Rage::Telemetry::Instrumenter
  instrument "controller.action.process", with: :instrument_controllers

  def self.instrument_controllers(context)
    start = Time.now.to_i
    yield
    puts "Completed #{context[:name]} in #{Time.now.to_i - start}"
  end
end
```


rsamoilov:
The PR has too many tangentially related changes in it. Additionally, the Telemetry approach will be redesigned.

PR #198, opened 2026-01-19
formatted: Fix connection pool spec syntax and unused variables

cuneyter:
## Description:

This PR performs minor cleanups in spec/ext/active_record/connection_pool_spec.rb:
- Removes an unused variable assignment (conn) in the "correctly releases connections" test block.
- Updates the establish_connection call to explicitly convert ENV["TEST_PG_URL"] to a string, improving code clarity and consistency.


rsamoilov (approved):
👍

PR #221, opened 2026-02-26
Shared Components in the @auth tag

### LINKED ISSUE #122: [OpenAPI] Shared Components in the `@auth` tag
### Description 

Currently, Rage supports specifying security schemes for an application using the [`@auth`](https://github.com/rage-rb/rage/wiki/OpenAPI-Guide#authentication) tag.

The tag supports two forms:

- using the default scheme:
```ruby
# @auth authenticate_user
```

- using a custom scheme inlined into the tag:

```ruby
# @auth authenticate_by_token
#   type: apiKey
#   in: header
#   name: X-API-Key
```

Unless it is something really simple and straightforward, we recommend avoiding inlining YAML into tags. Thus, it makes sense to update the `@auth` tag to accept a [shared reference](https://github.com/rage-rb/rage/wiki/OpenAPI-Guide#shared-references). For example:

```ruby
# @auth authenticate_by_token #/components/securitySchemes/ApiKeyAuth
```

Such tag would apply the `ApiKeyAuth` security scheme to all necessary controller actions.  

### Tips

- Check [OpenAPI Authentication Guide](https://swagger.io/docs/specification/v3_0/authentication/).
- The starting point for the code that generates OpenAPI documentation is [Rage::OpenAPI::Builder](https://github.com/rage-rb/rage/blob/master/lib/rage/openapi/builder.rb). The changes, however, will need to be made in [Rage::OpenAPI::Parser](https://github.com/rage-rb/rage/blob/master/lib/rage/openapi/parser.rb) and [Rage::OpenAPI::Converter](https://github.com/rage-rb/rage/blob/master/lib/rage/openapi/converter.rb).
- Check the [overview doc](https://github.com/rage-rb/rage/blob/master/OVERVIEW.md) that shows how Rage's core components interact with each other.
- **Feel free to ask any questions or request help in the comments below!**

Piyush-Goenka:
This PR adds support for using shared OpenAPI security scheme references in `@auth` tags.

New supported syntax:

```
# @auth authenticate_by_token #/components/securitySchemes/ApiKeyAuth
```

This allows `@auth` to reuse security schemes from shared components instead of requiring inline YAML definitions.

Closes #122.

1. Parser updates `lib/rage/openapi/parser.rb`

- Extended `@auth` parsing so the second token can be either:
- a normal custom name (existing behavior), or
- a shared reference path `#/components/securitySchemes/...`

 2. Converter updates `lib/rage/openapi/converter.rb`

-  If auth is from shared reference, just use its name and do not create a new scheme.
-  If auth is normal/old style, keep old behavior and create scheme as before.

3. Tests `spec/openapi/builder/auth_spec.rb`

- Added test coverage for new behavior:

4. Tests integration 

- Added the integration tests to really visualize how will the change work when someone is using this framework.
- Added the `sessions-controller.rb` to test for the exact auth tag behaviour in shared reference
- These integration test have been changed in this commit "Added integration tests"
Screenshot 2026-02-27 at 9 36 15 PM

### DIFF
--- a/lib/rage/openapi/parser.rb
+++ b/lib/rage/openapi/parser.rb
@@ -215,9 +215,18 @@ class Rage::OpenAPI::Parser
   end
 
   def parse_auth_tag(method:, name_or_ref:, children:, comment:)
+    if method&.start_with?("#/components/securitySchemes/")
+      Rage::OpenAPI.__log_warn "invalid `@auth` shared reference syntax detected at #{location_msg(comment)}; use `@auth <before_action> #{method}` and remove child definitions"
+      return
+    end
+
     if name_or_ref&.start_with?("#/components")
       ref = parse_auth_shared_reference(name_or_ref, comment)
       return if ref.nil?
+
+      if children.any?
+        Rage::OpenAPI.__log_warn "ignored child `@auth` definition detected at #{location_msg(comment)}; shared security scheme references do not support child definitions"
+      end
 
       return {
         method:,
--- a/spec/openapi/builder/auth_spec.rb
+++ b/spec/openapi/builder/auth_spec.rb
@@ -215,6 +215,47 @@ RSpec.describe Rage::OpenAPI::Builder do
       end
     end
 
+    context "with security scheme reference and child definition" do
+      before do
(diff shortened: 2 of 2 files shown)


rsamoilov (approved):
👍👍

Issue #215, opened 2026-02-24
[Deferred] Configurable retry options


rsamoilov:
### Description

Currently, `Rage::Deferred` automatically retries failed tasks up to 5 times using exponential backoff. However, this retry count is hardcoded and cannot be customized. Additionally, there is no way to specify which exceptions should trigger a retry.

The goal of this issue is to allow users to configure retry behavior per task:

1. **Custom retry count** - allow specifying the number of retry attempts (instead of the hardcoded 5).
2. **Retryable exceptions** - allow specifying a list of exception classes that should trigger a retry. All other exceptions would not be retried.

For example:

```ruby
class SyncWithExternalService
  include Rage::Deferred::Task

  retry_on Net::OpenTimeout, Net::ReadTimeout, attempts: 10
  # or
  retry_on attempts: 3

  def perform(user_id:)
    # ...
  end
end
```

### Tips

* Check the [Deferred docs](https://rage-rb.dev/docs/deferred) to understand how `Rage::Deferred` works.
* Look at the existing retry logic in the codebase to understand where the hardcoded `5` is used and how exponential backoff is applied.
* Consider what the default behavior should be when `retry_on` is not specified (the current behavior of 5 retries on all exceptions should be preserved for backward compatibility).
* Keep scope small: per-task config only; global config can be a future issue.
* Check the [architecture doc](https://github.com/rage-rb/rage/blob/main/ARCHITECTURE.md) that shows how Rage's core components interact with each other and outlines the design principles.
* **Feel free to ask any questions or request help in the comments below!**

PR #225, opened 2026-02-28
[Deferred] Add configurable retry options 

Digvijay-x1:
[Deferred] Add configurable retry options with max_retries and retry_interval
- Add `max_retries` class method for simple max attempt configuration
- Add `retry_interval(exception, attempt:)` as an overridable class method for advanced per-exception retry control:
  - Return an Integer to retry in N seconds
  - Return super for default exponential backoff
  - Return false/nil to abort retries
- Update `__should_retry?` and `__next_retry_in` to delegate to retry_interval
- Pass exception to `__next_retry_in` in the queue for per-exception delay decisions
  
Fixes: #215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Added
 
+- [Deferred] Support configurable retry options with `max_retries` and `retry_interval` by [@Digvijay](https://github.com/Digvijay-x1) (#215).
 - [Cable] Add support for `stop_stream_from` and `stop_stream_for`  by [@Digvijay](https://github.com/Digvijay-x1) (#217).
 
 ## [1.21.1] - 2026-02-27
--- a/lib/rage/deferred/metadata.rb
+++ b/lib/rage/deferred/metadata.rb
@@ -27,7 +27,7 @@ class Rage::Deferred::Metadata
     # @return [Boolean] `true` if a failure will schedule another attempt, `false` otherwise
     def will_retry?
       task = Rage::Deferred::Context.get_task(context)
-      task.__should_retry?(attempts)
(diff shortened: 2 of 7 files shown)


rsamoilov:
> The issue description says "Retryable exceptions - allow specifying a list of exception classes that should trigger a retry. All other exceptions would not be retried."

I’m really sorry - that was very bad wording on my end. 

I’ve also never used Active Job, so I didn’t realise that `retry_on` specifies an exclusive list of exceptions where everything else gets propagated to the backend.

`Rage::Deferred` is an execution layer itself, not an abstraction over a backend. When a task fails with an unexpected exception, there’s nowhere for it to propagate to - it’s either gets retried or it’s lost. So treating the exception list as exclusive would mean that unexpected errors silently fail. 

The behaviour that makes more sense for `Rage::Deferred` would be: exceptions on the list get retried with the custom retry count, and all other exceptions fall back to the default retry behaviour. The only thing is that considering how `retry_on` works in Active Job, a different behaviour in Rage might be surprising. 

Have you explored the hooks Sidekiq provides?

Digvijay-x1:
I looked into Sidekiq's error handling hooks, they have `sidekiq_options retry: N` for per-job config, `sidekiq_retry_in` for custom backoff logic, and `sidekiq_retries_exhausted` for a death callback. It's a solid system, but most of it is built around having a dead set and a web UI for manual retries, which doesn't really seems apply here.
 
 I think sticking with `retry_on` but fixing the fallback behavior is the simpler path for now. I've updated `__should_retry?` so that:

- Matching exceptions use the custom attempts count
- Non-matching exceptions fall back to MAX_ATTEMPTS (5)
- Default behavior stays the same

Let me know what you think!

Issue #216, opened 2026-02-24
[Cookies] Support signed cookies


rsamoilov:
### Description

Currently, Rage supports encrypted cookies via `cookies.encrypted`, which uses libsodium under the hood to both encrypt and authenticate cookie values. However, there is no support for signed cookies - cookies that are authenticated (tamper-proof) but not encrypted (the value is readable).

Signed cookies are useful when you want to ensure a cookie hasn't been tampered with, but the value itself doesn't need to be secret. This is a common pattern in Rails via `cookies.signed`.

The goal of this issue is to expose a `cookies.signed` API that uses libsodium's message authentication to sign and verify cookie values without encrypting them.

For example:

```ruby
# Setting a signed cookie
cookies.signed[:user_id] = 123

# Reading a signed cookie (returns nil if tampered with)
cookies.signed[:user_id] # => 123
```

### Tips

* Look at how `cookies.encrypted` is currently implemented in the codebase - the signed cookie implementation will follow a similar pattern but use authentication (signing) instead of encryption.
* Libsodium provides message authentication primitives that can be used for signing. Look into the [rbnacl](https://github.com/RubyCrypto/rbnacl) docs.
* Check the [architecture doc](https://github.com/rage-rb/rage/blob/main/ARCHITECTURE.md) that shows how Rage's core components interact with each other and outlines the design principles.
* **Feel free to ask any questions or request help in the comments below!**

PR #226, opened 2026-03-04
Add signed cookies

rfronczyk:
This PR adds support for signed cookies via cookies.signed in Rage.
 
 Closes: #216 

  Why

  Rage already supports encrypted cookies, but there was no option for integrity-only cookies.
  Signed cookies are useful when data can be readable but must be protected from tampering.

  What changed

  • Added cookies.signed API to Rage::Cookies.
  • Implemented SignedJar using rbnacl message authentication (RbNaCl::HMAC::SHA512256).
  • Added verification flow that returns nil for malformed/tampered cookies.
  • Added fallback key verification support for key rotation.
  • Added/updated specs to cover:
    • reading valid signed cookies,
    • writing signed cookies,
    • malformed payload handling,
    • tampering detection,
    • fallback key verification,
    • cookies.signed.permanent behavior.

  Cookie format

  Signed cookies use:
  `<base64_urlsafe_payload>.<base64_urlsafe_signature>`
  • `.` is the separator between payload and signature  as `.`is not part Base64 URL-safe characters list.
  • Payload is Base64 URL-safe encoded first.
  • Signature is computed over the encoded payload.
  • On read, Rage splits by `.`, verifies signature, and returns nil when format or signature is invalid.

  This is my first contribution to Rage, so feedback is very welcome.

### DIFF
--- a/lib/rage/cookies.rb
+++ b/lib/rage/cookies.rb
@@ -381,7 +381,7 @@ class Rage::Cookies
       def dump(value)
         encoded_value = Base64.urlsafe_encode64(value.to_s)
         signed_payload = signed_payload_for(VERSION, encoded_value)
-        "#{VERSION}#{SEPARATOR}#{encoded_value}#{SEPARATOR}#{digest_for(signed_payload, primary_signer)}"
+        "#{signed_payload}#{SEPARATOR}#{digest_for(signed_payload, primary_signer)}"
       end
 
       private


rfronczyk:
Hey @rsamoilov. I implemented your comments and fixed the rubocop complaints. Please take a look again when you have a chance.

PR #228, opened 2026-03-06
OpenAPI: support optional attributes and Array<> syntax in inline sch…

### LINKED ISSUE #227: [OpenAPI] Inline schema improvements: optional attributes and `Array<>` syntax
### Description

Inline YAML schemas used in `@request` and `@response` tags currently have two limitations:

#### 1. No support for optional attributes

When using an inline schema like `@response { id: Integer, name: String, email: String }`, all attributes are treated as non-required in the generated OpenAPI specification. There is no way to distinguish between required and optional attributes.

The `@param` tag already supports the `?` suffix to mark optional parameters. The same convention should work for inline schema attributes. Attributes **without** the `?` suffix should be marked as `required`, while attributes **with** the suffix should be non-required.

For example:

```ruby
class Api::V1::UsersController < ApplicationController
  # @request { name?: String, email: String, password: String }
  def create
  end
end
```

Should generate:

```yaml
schema:
  type: object
  required:
    - email
    - password
  properties:
    name:
      type: string
    email:
      type: string
    password:
      type: string
```

Note that `name` is still included in `properties` (without the `?`), but is excluded from the `required` list.

#### 2. No `Array<>` syntax for inline schema values

The `Array<>` syntax is already supported for top-level references (e.g. `@response Array<UserResource>`), and the `[]` syntax works inside inline schemas (e.g. `{ errors: [String] }`). However, `Array<>` is not supported inside inline schemas.

The following should work equivalently:

```ruby
# Currently supported
# @response 422 { errors: [String] }

# Should also be supported
# @response 422 { errors: Array<String> }
```

Both should generate:

```yaml
schema:
  type: object
  properties:
    errors:
      type: array
      items:
        type: string
```

### Tips

* Check the [OpenAPI docs](https://rage-rb.dev/docs/openapi) to understand how inline schemas, `@request`, and `@response` tags work.
* Look at how the inline schema parser currently handles the `[]` syntax - extending it to also accept `Array<>` should follow a similar pattern.
* The [OpenAPI specification on required properties](https://swagger.io/docs/specification/v3_0/data-models/data-types/#required-properties) describes how `required` works in object schemas.
* Check the [architecture doc](https://github.com/rage-rb/rage/blob/main/ARCHITECTURE.md) that shows how Rage's core components interact with each other and outlines the design principles.
* **Feel free to ask any questions or request help in the comments below!**

ayushman1210:
**Description**
 This PR addresses two limitations in the inline YAML schema parser used for @request and @response tags, as described in the issue 
1. **Optional Attributes (? suffix)**: Inline schemas now support marked optional attributes using the ? suffix (e.g., name?: String).
- Attributes without the suffix are now automatically added to the required array in the generated OpenAPI specification.
- Attributes with the suffix are treated as non-required and are excluded from the required list

2.**Array<> Syntax Support**: Added support for the Array<Type> syntax inside inline schemas to provide a consistent alternative to the unary array [Type] syntax.

- Example: { errors: Array<String> } now correctly generates an array of strings 

Changes
**File** --> [lib/rage/openapi/parsers/yaml.rb](url)

- Updated __parse  to handle key stripping and required array generation.
- Updated  type_to_spec to recognize and parse the Array<...> string format.

**File** --> [spec/openapi/parsers/yaml_spec.rb ](url)

- Added new test cases for optional/required attribute mixing and nested Array<> usage.
- Updated existing expectations to include the required property where applicable.

**Verification**

1. Added new RSpec tests covering mixed attribute optionality and varied Array<> types (String, Integer, Hash).
2. Verified that all existing tests pass with the adjusted requirements.
3. Manual verification of the parser logic with a standalone test script.

Fixes #227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 ## [Unreleased]
+
 ### Added
 - [Deferred] Add tests for log context capture and backward-compatible restore by [@jsxs0](https://github.com/jsxs0) (#274).
 - [OpenAPI] Add support for per-endpoint OAuth2/OpenID scopes via `@auth_scope` tag by [@Piyush-Goenka](https://github.com/Piyush-Goenka) (#272).
 - Reuse `define_dynamic_method` and `define_maybe_yield` methods in `RageController::API` from `Rage::Internal` by [@numice](https://github.com/numice) (#273).
 - Add the `form_actions` router configuration (#278).
 - [Deferred] Add native periodic task scheduling with multi-process leader election via `File#flock` by [@Abishekcs](https://github.com/Abishekcs) (#233).
+- [OpenAPI] Support optional attributes and `Array<>` syntax by [@ayushman1210](https://github.com/ayushman1210) (#228).
 
 ### Fixed
 
--- a/lib/rage/openapi/openapi.rb
+++ b/lib/rage/openapi/openapi.rb
@@ -120,7 +120,7 @@ module Rage::OpenAPI
 
   # @private
   def self.__try_parse_collection(str)
-    if str =~ /^Array<([\w\s:\(\)]+)>$/ || str =~ /^\[([\w\s:\(\)]+)\]$/
+    if str =~ /^Array<([\w\s:\(\),]+)>$/ || str =~ /^\[([\w\s:\(\),]+)\]$/
       [true, $1]
     else
       [false, str]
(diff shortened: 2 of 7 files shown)


rsamoilov (commented):
Hy @ayushman1210 

Left one small comment. Otherwise, looks good 👍

rsamoilov (commented):
Hi @ayushman1210 

This looks good. The only thing is that this doesn't seem to support enums. For example, the `[IN_PROGRESS, COMPLETED]` expression creates an enum, but `Array<IN_PROGRESS, COMPLETED>` is not parsed correctly.

rsamoilov (approved):
Thank you for your contribution, @ayushman1210 !

PR #252, opened 2026-03-30
SQL Adapter for Background Tasks

UtkarshDixit-97:
This is a Sample PR of a very basic implementation of SQL based WAL storage for deffered task.
chart

* Architecture overview - Implemented a lightweight, cloud-native SQL adapter for Rage::Deferred to replace the disk-based WAL. The adapter persists deferred tasks durably in a SQL database while keeping task execution local to the same process.
* Key components -
  - Rage::Deferred::Backends::SQL – The main adapter implementing add, remove, and pending_tasks
  - Rage::Deferred::Queue – Uses the backend for persistence and replay
  - Automatic table creation (rage_deferred_tasks) with task_id (UUID), serialized_task, publish_at, and status
  - Configuration integration via Rage.configure { config.deferred.backend = :sql }

* Implementation strategy -
  - Database access during framework boot so we can create rage_deferred_tasks table(later on checking pending_tasks[], which will return pending or failed tasks )
  - For any task which is referred we will store log in db table before enqueing it in memory
  - On task is scheduled and completed (remove), the row is deleted instead of marked for deletion

* Technologies or libraries involved
  - ActiveRecord (for connection management and schema operations)
  - SQLite3 (currently tested) / PostgreSQL/MySQL (future implementation if proposal is accepted in GSOC)
  - Rage’s Fiber Scheduler + custom ConnectionPool extension
  - SecureRandom.uuid for task IDs

For the Application to switch to sql based log, create a file --> config/initializers/rage.rb

```
Rage.configure do
  # Other configurations if any...

  config.deferred.backend = :sql
  # Optional: custom table name
  # config.deferred.backend = :sql, table_name: "rage_deferred_tasks"
end
```

### DIFF
--- a/lib/rage/configuration.rb
+++ b/lib/rage/configuration.rb
@@ -702,10 +702,13 @@ class Rage::Configuration
       when :disk
         @backend_options = parse_disk_backend_options(opts)
         Rage::Deferred::Backends::Disk
+      when :sql
+        @backend_options = parse_sql_backend_options(opts)
+        Rage::Deferred::Backends::SQL
       when nil
         Rage::Deferred::Backends::Nil
       else
-        raise ArgumentError, "unsupported backend value; supported keys are `:disk` and `nil`"
+        raise ArgumentError, "unsupported backend value; supported keys are `:disk`, `:sql`, and `nil`"
       end
     end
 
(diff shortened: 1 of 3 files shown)


Abishekcs:
Reason: Google Summer of Code POC.

PR #257, opened 2026-04-01
feat(deferred): Improve default retry strategy

xingzihai:
## Summary

This PR improves the default retry strategy for `Rage::Deferred` tasks as described in #251.

### Changes

1. **Increased MAX_ATTEMPTS** from 5 to 15
   - With 5 retries, tasks exhausted all attempts in ~2.7 minutes
   - With 15 retries, tasks have ~2 days before exhausting all attempts
   - This gives developers more time to discover and fix underlying issues

2. **Changed backoff formula** from exponential to polynomial
   - Old: `rand(5 * 2**attempt) + 1` (exponential growth)
   - New: `(attempt**4) + 10 + (rand(15) * attempt)` (polynomial growth)
   - This spaces retries further apart with each attempt

### Retry timing comparison

| Attempt | Old (avg) | New (avg) |
|---------|-----------|-----------|
| 1       | ~3s       | ~18s      |
| 2       | ~6s       | ~41s      |
| 3       | ~11s      | ~2m       |
| 4       | ~21s      | ~5m       |
| 5       | ~41s      | ~11m      |
| 10      | -         | ~2.8h     |
| 15      | -         | ~14.1h    |

### Total retry time

- Old: ~2.7 minutes (5 retries)
- New: ~2 days (15 retries)

### Testing

Updated spec tests to verify the new backoff intervals:
- `__next_retry_in(0, nil)` returns between 11-25 seconds
- `__next_retry_in(1, nil)` returns between 26-54 seconds
- `__next_retry_in(15, nil)` returns between 50635-50845 seconds (~14h)
- `__next_retry_in(16, nil)` returns nil (max exceeded)

Closes #251

### DIFF
--- a/lib/rage/deferred/task.rb
+++ b/lib/rage/deferred/task.rb
@@ -31,11 +31,8 @@
 # ```
 #
 module Rage::Deferred::Task
-  MAX_ATTEMPTS = 5
+  MAX_ATTEMPTS = 15
   private_constant :MAX_ATTEMPTS
-
-  BACKOFF_INTERVAL = 5
-  private_constant :BACKOFF_INTERVAL
 
   # @private
   CONTEXT_KEY = :__rage_deferred_execution_context
@@ -191,7 +188,9 @@ module Rage::Deferred::Task
 
     # @private
     def __default_backoff(attempt)
-      rand(BACKOFF_INTERVAL * 2**attempt) + 1
+      # Use 1-indexed attempt for the formula as per issue #251
+      retry_attempt = attempt + 1
+      (retry_attempt**4) + 10 + (rand(15) * retry_attempt)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


anuj-pal27 (changes requested):
I think there are two changes needed to align with issue #251 and maintainer feedback:

* **Set `MAX_ATTEMPTS` to 20 (not 15).**
  In the issue discussion, the direction moved toward 20 to better cover real-world outages (especially weekend/longer incidents). 15 improves things, but 20 gives a much safer retry window by default, which is the main goal of #251.
* **Don’t increment `attempt` in `__default_backoff`:** Using `attempt + 1` shifts the retry timing and skips the intended first tier (~18s avg), making retry 1 behave like retry 2.

We should keep the formula exactly as specified:

```ruby
def __default_backoff(attempt)
  (attempt**4) + 10 + (rand(15) * attempt)
end
```
@rsamoilov — is this understanding correct?

rsamoilov:
Implemented in #271.

PR #268, opened 2026-04-08
Fix SystemStackError in Alba parser with circular associations

pratyush07-hub:
## Description
The Alba parser in **Rage::OpenAPI** currently lacks a recursion guard for associations. When two resources reference each other circularly (e.g., **UserResource ↔ PostResource**), the parser enters an infinite loop while attempting to recursively nest the schemas, eventually resulting in a **SystemStackError**.

This PR introduces a dynamic schema registry and switches the Alba parser to use **$ref** pointers for circular associations. This prevents infinite recursion while preserving full schema information and aligning with OpenAPI best practices.

----
## Changes

- Added `__schema_registry` (a **Hash**) to Rage::OpenAPI to store dynamically discovered schemas.
- Updated **Rage::OpenAPI::Parsers::Ext::Alba**:

  - __parse_nested now returns a $ref pointer (e.g., { "**$ref":"#/components/schemas/UserResource**" }) when a circular reference is detected.
  - parse registers the full clean schema if it was referenced via **$ref** during the parsing process.

- Updated **Rage::OpenAPI::Converter** to merge these dynamic schemas into **components/schemas**.

----
## Testing
Automated Tests

- Added `spec/openapi/parsers/ext/alba_recursion_spec.rb`
- Commands:
``` ruby
bundle exec rspec spec/openapi/parsers/ext/alba_recursion_spec.rb
bundle exec rspec spec/openapi/parsers/ext/alba_spec.rb spec/openapi/parsers/ext/alba_fallback_spec.rb
```

----
Closes: #265

### DIFF
--- a/lib/rage/openapi/builder.rb
+++ b/lib/rage/openapi/builder.rb
@@ -21,6 +21,7 @@ class Rage::OpenAPI::Builder
   end
 
   def run
+    Rage::OpenAPI.__reset_data_cache
     parser = Rage::OpenAPI::Parser.new
 
     @routes.each do |controller, routes|
--- a/lib/rage/openapi/builder.rb
+++ b/lib/rage/openapi/builder.rb
@@ -21,9 +21,7 @@ class Rage::OpenAPI::Builder
   end
 
   def run
-    # reset the data cache at the beginning of each build to prevent data leakage between builds in long-running processes
-    Rage::OpenAPI.__reset_data_cache
-    parser = Rage::OpenAPI::Parser.new
+    parser = Rage::OpenAPI::Parser.new(@nodes)
 
     @routes.each do |controller, routes|
       next if skip_controller?(controller)
--- a/lib/rage/openapi/builder.rb
+++ b/lib/rage/openapi/builder.rb
@@ -21,6 +21,7 @@ class Rage::OpenAPI::Builder
   end
 
   def run
+    # reset the data cache at the beginning of each build to prevent data leakage between builds in long-running processes
     Rage::OpenAPI.__reset_data_cache
     parser = Rage::OpenAPI::Parser.new
 
--- a/lib/rage/openapi/parsers/ext/alba.rb
+++ b/lib/rage/openapi/parsers/ext/alba.rb
@@ -16,10 +16,27 @@ class Rage::OpenAPI::Parsers::Ext::Alba
   end
 
   def parse(klass_str)
-    __parse(klass_str).build_schema
+    _, raw_klass_str = Rage::OpenAPI.__try_parse_collection(klass_str)
+    visitor = __parse(klass_str)
+
+    if Rage::OpenAPI.__schema_registry.key?(raw_klass_str)
+      clean = { "type" => "object" }
+      clean["properties"] = visitor.schema if visitor.schema.any?
+      Rage::OpenAPI.__schema_registry[raw_klass_str] = clean
+    end
+
+    visitor.build_schema
   end
 
   def __parse_nested(klass_str)
+    is_collection, raw_klass_str = Rage::OpenAPI.__try_parse_collection(klass_str)
+
+    if @parsing_stack.include?(raw_klass_str)
+      Rage::OpenAPI.__schema_registry[raw_klass_str] ||= nil
+      ref = { "$ref" => "#/components/schemas/#{raw_klass_str}" }
+      return is_collection ? { "type" => "array", "items" => ref } : ref
+    end
+
     __parse(klass_str).tap { |visitor|
       visitor.root_key = visitor.root_key_for_collection = visitor.root_key_proc = visitor.key_transformer = nil
     }.build_schema
--- a/lib/rage/openapi/parsers/ext/alba.rb
+++ b/lib/rage/openapi/parsers/ext/alba.rb
@@ -48,7 +48,6 @@ class Rage::OpenAPI::Parsers::Ext::Alba
     if @parsing_stack.include?(klass_str)
       return Visitor.new(self, is_collection)
     end
-    
     @parsing_stack.add(klass_str)
 
     klass = @namespace.const_get(klass_str)
--- a/lib/rage/openapi/parsers/ext/alba.rb
+++ b/lib/rage/openapi/parsers/ext/alba.rb
@@ -45,6 +45,8 @@ class Rage::OpenAPI::Parsers::Ext::Alba
   def __parse(klass_str)
     is_collection, klass_str = Rage::OpenAPI.__try_parse_collection(klass_str)
 
+    # return an empty visitor if we're already parsing this class;
+    # this serves as a recursion guard, specifically for the inheritance logic in `Visitor#visit_class_node`
     if @parsing_stack.include?(klass_str)
       return Visitor.new(self, is_collection)
     end
--- a/lib/rage/openapi/parsers/ext/alba.rb
+++ b/lib/rage/openapi/parsers/ext/alba.rb
@@ -5,6 +5,7 @@ class Rage::OpenAPI::Parsers::Ext::Alba
 
   def initialize(namespace: Object, **)
     @namespace = namespace
+    @parsing_stack = Set.new
   end
 
   def known_definition?(str)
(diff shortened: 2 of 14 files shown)


lib/rage/openapi/parsers/ext/alba.rb
```diff
@@ -15,10 +16,27 @@ def known_definition?(str)
   end
 
   def parse(klass_str)
-    __parse(klass_str).build_schema
+    _, raw_klass_str = Rage::OpenAPI.__try_parse_collection(klass_str)
```
rsamoilov, line 19: This line is currently not covered with tests. Try changing it to `raw_klass_str = klass_str` to see the tests pass.

spec/openapi/parsers/ext/alba_recursion_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+
+RSpec.describe Rage::OpenAPI::Parsers::Ext::Alba do
+  include_context "mocked_classes"
+
+  subject { described_class.new.parse(resource) }
+
+  let(:resource) { "UserResource" }
+
+  # Clear the schema registry between tests so $ref registrations don't leak
+  before { Rage::OpenAPI.__schema_registry.clear }
```
rsamoilov: Is it still needed considering `__reset_data_cache` is called in the `spec_helper.rb`?

lib/rage/openapi/parsers/ext/alba.rb
```diff
@@ -27,13 +45,20 @@ def __parse_nested(klass_str)
   def __parse(klass_str)
     is_collection, klass_str = Rage::OpenAPI.__try_parse_collection(klass_str)
 
+    if @parsing_stack.include?(klass_str)
```
rsamoilov, line 50: Does this check exist to correctly support inheritance? Can you please add a comment explaining this block?

rsamoilov (commented):
Hi @pratyush07-hub ,

I left several minor comments.

Apart from that, I like that you use `$ref` for circular refs. Tests also look good 👍

lib/rage/openapi/parsers/ext/alba.rb
pratyush07-hub: I've fixed this and added a test case to cover this scenario in `spec/openapi/parsers/ext/alba_recursion_spec.rb`.

spec/openapi/parsers/ext/alba_recursion_spec.rb
pratyush07-hub: You're right! I checked `spec/spec_helper.rb` and I have removed the redundant before { **Rage::OpenAPI.__schema_registry.clear** } from the spec and verified that all tests still pass.

lib/rage/openapi/parsers/ext/alba.rb
pratyush07-hub: Yes, this check acts as a recursion guard. It's particularly important for the inheritance logic in `Visitor#visit_class_node` . Without this, circular inheritance or repetitive parsing of the same class in a hierarchy could lead to a stack overflow.

I've added a comment explaining this block as requested.

pratyush07-hub:
@rsamoilov , I have addressed all the minor comments you pointed out and made the necessary changes. Could you please take another look?

lib/rage/openapi/builder.rb
```diff
@@ -21,6 +21,7 @@ def initialize(namespace: nil)
   end
 
   def run
+    Rage::OpenAPI.__reset_data_cache
```
rsamoilov, line 25: Why do we need this call here? Tests pass without it.

rsamoilov (commented):
Hey @pratyush07-hub , this looks ready to merge. Just have one question about a line I might've missed the first time.

lib/rage/openapi/builder.rb
```diff
@@ -21,6 +21,7 @@ def initialize(namespace: nil)
   end
 
   def run
+    # reset the data cache at the beginning of each build to prevent data leakage between builds in long-running processes
```
rsamoilov, line 24: > data leakage between builds in long-running processes

So I don’t think this can happen - `run` is called only once by `Rage::OpenAPI.application` and then cached.

lib/rage/openapi/builder.rb
pratyush07-hub: > Why do we need this call here? Tests pass without it.

I've addressed your comment by adding an explanatory comment to `lib/rage/openapi/builder.rb` regarding the necessity of the cache reset, and I've added a verification test in `spec/openapi/builder/cache_leakage_spec.rb` to ensure it works as expected.

lib/rage/openapi/builder.rb
pratyush07-hub: > > data leakage between builds in long-running processes
> 
> So I don’t think this can happen - `run` is called only once by `Rage::OpenAPI.application` and then cached.

You're right that for a single namespace in production, `run` is typically only called once. However, in development (with reloading) or when multiple OpenAPI apps are mounted in the same process with different namespaces, `run` can be called several times. Without clearing the `__schema_registry` at the start of `run`, we risk leaking circular association schemas from previous builds or other namespaces.

Because clearing the entire cache also invalidates the final spec cache for other namespaces, should I refine this to only clear the `__schema_registry` specifically? This would prevent data leakage between builds while preserving the performance benefits of caching for other mounted OpenAPI apps.

rsamoilov: > Without clearing the __schema_registry at the start of run, we risk leaking circular association schemas from previous builds or other namespaces

Can you please share the steps to reproduce this issue? The cache should be cleared automatically in development - https://github.com/rage-rb/rage/blob/main/lib/rage/code_loader.rb#L88.

pratyush07-hub: > Can you please share the steps to reproduce this issue? The cache should be cleared automatically in development - https://github.com/rage-rb/rage/blob/main/lib/rage/code_loader.rb#L88.

### Steps to Reproduce:

**Create a script `repro_leakage.rb`**:
```ruby
# frozen_string_literal: true

$LOAD_PATH.unshift(File.expand_path("lib", __dir__))
require "rage"

# 1. Simulate Build 1 (e.g., Namespace V1)
# Circular associations found during this build are stored in the global registry.
Rage::OpenAPI.__reset_data_cache
Rage::OpenAPI.__schema_registry["UserResource"] = { "type" => "object", "note" => "V1 Circular Schema" }

puts "Registry keys after V1: #{Rage::OpenAPI.__schema_registry.keys}"

# 2. Simulate Build 2 (e.g., Namespace V2)
# In production or a test suite, Build 2 happens in the same process.
# WITHOUT a reset at the start of Builder#run, the registry still contains V1's data.
if Rage::OpenAPI.__schema_registry.key?("UserResource")
  puts "!! LEAKAGE DETECTED: Build 2 sees 'UserResource' from V1 !!"
  puts "This can lead to V2 incorrectly using V1's schema or skipping its own circularity checks for a class with the same name."
end
```

**Run the script**: `ruby repro_leakage.rb`

`CodeLoader` only resets the cache in **development mode** during a **code reload**. In **production**, if you host multiple API versions (V1, V2) in the same process, `CodeLoader` will not run between requests for different versions. In **tests**, multiple builds are often triggered in a single RSpec run.

rsamoilov: Ahh, so that's for namespaces isolation.

That's a really great side effect you've found! But in this case, the approach to reset the cache is slightly wrong:

* The cache is fundamentally global, so we can't reliably isolate data from different OpenAPI namespaces in it. The fact that you have to reset it every time the spec has to be built means it's not the right tool for the task.
* The cache also stores the built specifications. Imagine there're two endpoints for two different namespaces. What will happen with this solution is that every request to build the specification for one of the namespaces will reset the cache for another namespace, effectively negating its value.

Consider storing the schema registry in the [root node](https://github.com/rage-rb/rage/blob/main/lib/rage/openapi/nodes/root.rb) instead.

pratyush07-hub:
Hi @rsamoilov , You were absolutely right, using a global singleton and resetting it during `Builder#run` was the wrong approach, as it unnecessarily invalidated cached specifications for other namespaces.

I have refactored the implementation to align with your recommendation by moving the schema registry directly into the `Rage::OpenAPI::Nodes::Root` instance. Since `Nodes::Root` is created per-build, this ensures complete namespace isolation without relying on the global cache. Consequently, I've removed the global cache reset from `Builder#run`, which now allows the framework to correctly preserve built specifications across builds while preventing one namespace from wiping another's data.

To support this architectural shift, I updated the parser hierarchy to propagate the `Root` node through the `Parser`, `Response`, and `Request` components down to the `Alba` parser. I've verified the fix with a new integration test proving the global registry remains empty and namespaces are fully isolated. All 298 examples in the OpenAPI test suite are now passing.

rsamoilov (approved):
Amazing fix!

PR #270, opened 2026-04-09
Support launching with rackup on Rack 3

pratyush07-hub:
## Description
Currently, Rage works correctly on Rack 2 but fails on Rack 3 due to the move of the handler registry to the rackup gem and the new **Rackup::Handler** namespace.
This PR adds support for launching Rage applications using the **rackup** command on Rack 3.

----
## Changes

- **Added Rage::Handler**: A new handler that delegates to the optimized **Rage::CLI** server logic, ensuring Rage-specific features like Fiber-concurrency are active.
- **Handler Registration**: Registered the new handler in both Rack::Handler (Rack 2) and Rackup::Handler (Rack 3) within `lib/rage-rb.rb`.
- **Discovery Support**: Added discovery files in `lib/rack/handler/rage.rb` and `lib/rackup/handler/rage.rb` to enable automatic identification of the Rage handler via the **-s rage** flag.

----
## Verification

- Verified the registration logic via a dedicated verification script.
- Confirmed that the rackup command correctly identifies the Rage handler when the rackup gem is present.

----
Closes #259


pratyush07-hub:
Hi @rsamoilov , This PR is now ready for your review. PTAL!

rsamoilov:
Hi @pratyush07-hub 

I can see the following error when trying to run `rackup` with your PR:

```
/Users/r/.gem/ruby/3.4.1/gems/rackup-2.3.1/lib/rackup/server.rb:269:in 'block in Rackup::Server.logging_middleware': undefined method 'name' for nil (NoMethodError)

          /CGI/.match?(server.server.name) || server.options[:quiet] ? nil : [Rack::CommonLogger, $stderr]
```

Am I doing something wrong?

Also, there's an existing Rack handler in Iodine - https://github.com/rage-rb/iodine/blob/master/lib/rack/handler/iodine.rb. Do you think these changes should go in there instead?

pratyush07-hub:
@rsamoilov , The **NoMethodError** occurred because the handler wasn't explicitly registered in the discovery files, causing it to be **nil** in some loading scenarios. I have fixed this by adding explicit registration to `lib/rack/handler/rage.rb` and `lib/rackup/handler/rage.rb`, and added a name method to **Rage::Handler**. 

Regarding the move to iodine, keeping the rage handler allows for better branding (**rackup-s rage**) and gives us direct control over Rage-specific startup logic. Furthermore, providing discovery files in rage makes the handler immediately available without requiring users to know about the internal relationship with iodine. 

Additionally, **rage-iodine** currently lacks Rackup (Rack 3) registration, so having it here ensures full support for the latest Rack versions. 
I have updated the PR with these changes.

PR #273, opened 2026-04-12
Reused define_dynamic_method and define_maybe_yield from Rage::Internal

numice:
Hi. As suggested earlier, I've created this pull request for reusing methods in `controller/api.rb` that are moved to `Internal`. 

This is my first ever pull request and not entirely sure if I did it right. So suggestions are welcome.

### DIFF
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require_relative "../internal"
-
 class RageController::API
   class << self
     # @private
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "../internal"
+
 class RageController::API
   class << self
     # @private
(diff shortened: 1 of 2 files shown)


lib/rage/controller/api.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "../internal"
```
rsamoilov: You don't need this call - the `Internal` module is [always loaded](https://github.com/rage-rb/rage/blob/b66473cf53a22066212cb930983b07f3bdc2c2bb/lib/rage-rb.rb#L202).

rsamoilov (commented):
Hi @numice , I've got one minor comment. Apart from that, looks good!

rsamoilov:
> This is my first ever pull request and not entirely sure if I did it right

Congratulations! You did it totally right.

rsamoilov:
Could you please also add a changelog entry to the [Unreleased section](https://github.com/rage-rb/rage/blob/main/CHANGELOG.md)?

numice:
@rsamoilov Thanks a lot for the feedback. I will fix it shortly.

numice:
I've updated it now.

rsamoilov (approved):
Looks great! Such consolidating changes are always the best. Thank you!

rsamoilov:
Hi @numice ,

Any chance you're up for working on another issue?

numice:
@rsamoilov Hi. Thank you for asking. It was great that my first ever pull request got merged. Yes, absolutely. I'm up for something within this kind of difficulty cause I'm still learning the rope.

rsamoilov:
What do you think about updating the [`head`](https://github.com/rage-rb/rage/blob/main/lib/rage/controller/api.rb#L532) method to accept a second argument with the response headers to set? It would mimic the way this method works [in Rails](https://api.rubyonrails.org/classes/ActionController/Head.html).

numice:
@rsamoilov That sounds good. I will work on it.

PR #274, opened 2026-04-14
Add tests for Deferred log context capture and backward-compatible restore

jsxs0:
### Summary

While working on the SSE context propagation tests in #267, I traced how log context flows across fiber boundaries in the Deferred system, specifically `Context.build` (capture) and `restore_log_info` in `Task` (restore). I noticed two paths without test coverage:

**1. `Context.build` log context capture**

The `"when rage_logger is set"` context verified `context[4]` (log tags) but not `context[5]` (log context). Added a test confirming `Fiber[:__rage_logger_context]` is captured, and fixed the `after` block to clean up both fiber-locals.

**2. `restore_log_info` backward compatibility**

`restore_log_info` has three branches: the current Array format, a legacy string format (older versions stored `request_id` as a plain string at index 4), and nil. The Array and nil branches were tested, but the legacy string path, which wraps the value into an array and defaults context to `{}`, had no coverage.

### Test plan

```
bundle exec rspec spec/deferred/context_spec.rb spec/deferred/task_spec.rb
58 examples, 0 failures
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## [Unreleased]
+### Added
+- [Deferred] Add tests for log context capture and backward-compatible restore by [@jsxs0](https://github.com/jsxs0) (#274).
 
 ### Fixed
 - [SSE] Ensure connection is closed for single-value SSE streams by [@jsxs0](https://github.com/jsxs0) (#264).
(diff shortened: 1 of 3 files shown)


rsamoilov (approved):
Amazing!

PR #275, opened 2026-04-14
Add centralized error reporting via Rage::Errors

### LINKED ISSUE #232: [Errors] Centralized error reporting interface
### Description

Rage has multiple independent components that catch and log exceptions internally — controllers, background tasks, Cable connections, Cable channels, event subscribers, and SSE streams. Currently, there is no way to tell the framework where to forward these exceptions beyond the built-in logger.

In practice, most applications use an external error tracking service (e.g. Sentry, Bugsnag, Rollbar). Today, the only way to get Rage-caught exceptions into these services is to manually rescue errors in each component and forward them yourself. This is repetitive and easy to miss.

The goal of `Rage::Errors` is to provide a centralized hook that lets users tell Rage: "when you catch an exception, also report it here". This is not a general-purpose error reporting abstraction or a replacement for the error tracking service's SDK — it's a way for the framework to forward exceptions it already catches.

#### Proposed interface

Register one or more reporters — any object that responds to `call`:

```ruby
# config/application.rb
Rage.configure do
  config.errors << SentryReporter.new
  config.errors << InternalMetricsReporter.new
end
```

Where a reporter looks like:

```ruby
class SentryReporter
  def call(exception, context: {})
    Sentry.capture_exception(exception, extra: context)
  end
end
```

Users should also be able to report errors manually using the `Rage::Errors` interface:

```ruby
Rage::Errors.report(exception)
```

#### Design considerations

These are open questions that should be addressed in the design proposal:

* **Logging behavior when a reporter is set.** The framework currently logs all caught exceptions. Should it continue to do so when a reporter is configured?
* **Multiple reporters.** The interface above supports multiple reporters. The implementation should iterate over all subscribers and call `call` on each. Consider what happens if one reporter raises an exception.
* **Duplicate errors.** Should the framework attempt to deduplicate errors before reporting?
* **Reporting unraised exceptions.** `Rage::Errors.report(SomeError.new("..."))` should work for manually created exceptions.
* **Reporting interface** Consider whether `call` should accept an optional severity/source argument.
* **Integration points.** Identify all the places in the codebase where Rage rescues exceptions (controllers, deferred tasks, Cable connections/channels, event subscribers, SSE streams) and ensure each one calls the reporting hook.

### Before You Start

This issue requires a design proposal before implementation. Please open a discussion or comment on this issue with a high-level plan that outlines:

1. The public API (confirming or adjusting the interface above).
2. How reporters will be registered and invoked.
3. The list of integration points in the codebase where reporting will be added.
4. How the edge cases above will be handled.

PRs without prior design discussion are unlikely to be accepted.

### Tips

* Search the codebase for `rescue` and `Rage.logger.error` to find all the places where Rage catches and logs exceptions — these are the integration points.
* Check the [architecture doc](https://github.com/rage-rb/rage/blob/main/ARCHITECTURE.md) that shows how Rage's core components interact with each other and outlines the design principles.
* For reference, see Rails' [`Rails.error` API](https://guides.rubyonrails.org/error_reporting.html). Note that `Rage::Errors` has a different goal — it's not about reducing boilerplate or providing a standard reporting interface, it's about letting the framework forward exceptions it already catches.
* **Feel free to ask any questions or request help in the comments below!**

Digvijay-x1:
Allow users to register error reporters that receive all framework-caught exceptions, so they can forward them to services like Sentry or Bugsnag without manually rescuing in every component.
```ruby
  Rage.configure do
    config.errors << SentryReporter.new
  end
  ```
Reporters can accept just the exception or an optional context hash. Signature detection uses Rage::Internal.build_arguments at registration time to keep the reporting path fast. Each reporter is called in its own rescue block so a broken reporter doesn't affect the rest. Duplicate prevention uses an instance variable on the exception (similar to Rails). Manually-created exceptions get a real backtrace via re-raise rather than set_backtrace(caller).
Hooked into all existing framework rescue sites: controllers, fiber wrapper, Cable connections/channels, Redis adapter, deferred tasks, event subscribers, SSE streams, and telemetry tracers.

Fixes #232

### CODE REVIEW

**Comment by rsamoilov** [comment]
Hey @Digvijay-x1 ,

I usually don't review in-progress PRs, so I haven't looked at this one yet. Anything I can help with? Do you want me to give it a look?

**Comment by Digvijay-x1** [comment]
> Hey @Digvijay-x1 ,
> 
> I usually don't review in-progress PRs, so I haven't looked at this one yet. Anything I can help with? Do you want me to give it a look?

Hey @rsamoilov,

Thanks for offering to help :)

Before requesting a review, I wanted to test whether all cases behave as expected. This took a bit longer since I had some college assignments last week.

From my testing so far, the core error reporting API appears to be working well. I set up a small Rage app with Sentry integration and verified that reporter registration, error forwarding, context propagation, duplicate prevention, and backtrace generation all behave correctly.

I’ve tested 5 out of the 9 call sites so far — including unhandled controller exceptions (covering both application.rb and fiber_wrapper.rb), deferred task failures, and direct Rage::Errors.report calls.

The remaining areas (events/subscriber, SSE, Cable, Redis adapter, telemetry) all route through the same Rage::Errors.report method, which is already working as expected.

I think the PR is in a good state for review now. Would appreciate your feedback whenever you get a chance!

**Review by rsamoilov — COMMENTED** [review]
Hey @Digvijay-x1 , the code looks good. I'll test the implementation more thoroughly and get back to you early next week.

File: lib/rage/errors.rb [hunk 1]
```diff
@@ -1,4 +1,72 @@
...
+    def report(exception, context: {})
+      return if @reporters.empty?
+      return if exception.instance_variable_defined?(:@_rage_error_reported)
+
+      ensure_backtrace(exception)
+
+      @reporters.length.times do |i|
+        __send__(:"__report_#{i}", exception, context)
+      rescue => e
+        Rage.logger.error("Error reporter #{@reporters[i].class} failed: #{e.class} (#{e.message})")
```
rsamoilov:
```suggestion
        Rage.logger.error("Error reporter #{@reporters[i].class} failed while reporting #{exception.class}: #{e.class} (#{e.message})")
```

rsamoilov (commented):
Hi @Digvijay-x1 ,

This is an amazing implementation. I love how minimal it is - this is exactly what I had in mind.

My main comment is that you're mixing two different concerns (configuration and reporting) inside `Rage::Errors`. 

`report` should be part of the module - that's OK. But configuration (`<<`) belongs to `Rage::Configuration` - you should have a separate namespace there that allows to add and remove error reporters. For example:

```ruby
Rage.configure do
  config.error_handlers << MyErrorHandler.new
end
```

Then, once a reporter is added, the configuration code should call a `@private` method on `Rage::Errors` to actually register the reporter.

rsamoilov:
Let's also add the reporter hook into [`Rage::PubSub::Adapters::Redis`](https://github.com/rage-rb/rage/blob/main/lib/rage/pubsub/adapters/redis.rb) (that's a new module that will replace `Rage::Cable::Adapters::Redis`) and add a shortcut to access `Rage::Errors` via `Rage.errors` to the [main module](https://github.com/rage-rb/rage/blob/main/lib/rage-rb.rb).

rsamoilov (approved):
Looks good! Can't wait to include this into the next release!

PR #277, opened 2026-04-18
feat(deferred): add periodic task scheduling

Abishekcs:
#233

## What this PR Does
Adds native recurring task scheduling to `Rage::Deferred`, allowing tasks to run on a fixed interval without relying on external tools like `cron` or third-party libraries.

## Usage
```ruby
Rage.configure do
  config.deferred.schedule do
    every 1.hour,   task: CleanupExpiredInvites
    every 1.minute, task: ResetCache
  end
end
```
Where `CleanupExpiredInvites` and `ResetCache` are standard `Rage::Deferred::Task` classes with a `perform` method.

## Implementation
Uses Iodine's `run_every` timer primitive. No new threads or processes the scheduler runs within the existing Iodine runtime. When a timer fires, it calls `task.enqueue` which goes through the normal Rage::Deferred path  WAL write, middleware, worker execution.

Schedule blocks are stored during configuration and evaluated at boot after all app constants are loaded, so task classes defined in `app/tasks/` are always available.

Leader election uses `File#flock` the same pattern used by `Rage::Cable`. Each worker tries a non-blocking exclusive lock on a shared file at boot. The winner registers timers, others stand by. When the leader dies, the OS releases the lock and the next spawned worker takes over.

## Design Decisions
- **Overlaps allowed by default** matches Sidekiq, Que, and GoodJob behavior.
- **No arguments to perform** scheduled tasks are stateless and discover their own data at runtime.
- **No polling loop** boot-time election is enough since Iodine always respawns dead workers.
- **Timers reset on leader change** no catch-up for missed intervals.

## Screenrecording:

`4 workers, 1 worker wins leader election and 2 task scheduled`

https://github.com/user-attachments/assets/a9610894-4d92-4bcf-8fe0-2235a2670429

`4 workers, 1 worker wins leader election and 2 task scheduled and leader worker dies after some time`

https://github.com/user-attachments/assets/4d688c52-40d6-468c-97a7-703a9a88585c

### DIFF
--- a/lib/rage/configuration.rb
+++ b/lib/rage/configuration.rb
@@ -715,6 +715,39 @@ class Rage::Configuration
     # @private
     def initialize
       @configured = false
+      @schedule_blocks = []
+    end
+
+    # Stores the scheduling block for later execution
+    def schedule(&block)
+      @schedule_blocks << block
+    end
+
+    # Evaluates all stored schedule blocks and returns the collected tasks.
+    # Called at boot time after all app constants are loaded.
+    def scheduled_tasks
+      @schedule_blocks.flat_map do |block|
+        dsl = ScheduleDSL.new
+        dsl.instance_eval(&block)
+        dsl.tasks
+      end
+    end
+
+    # @private
+    class ScheduleDSL
+      attr_reader :tasks
+
+      def initialize
+        @tasks = []
+      end
+
+      # Registers a task to run on a fixed interval (in seconds)
+      def every(interval, task:)
+        unless task.is_a?(Class) && task.include?(Rage::Deferred::Task)
+          raise ArgumentError, "#{task} must be a class that includes Rage::Deferred::Task"
+        end
+        @tasks << { interval:, task: }
+      end
     end
 
     # Returns the backend instance used by `Rage::Deferred`.
--- a/lib/rage/deferred/scheduler.rb
+++ b/lib/rage/deferred/scheduler.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class Rage::Deferred::Scheduler
+  LOCK_PATH = "/tmp/rage_deferred_scheduler.lock"
+
+  def self.start(tasks)
+    return if tasks.empty?
+
+    Rage::Internal.pick_a_worker(lock_path: LOCK_PATH) do
+      Rage.logger.info " Worker PID #{Process.pid} is managing scheduled tasks"
+      register_timers tasks
+    end
+  end
+
+  def self.register_timers(tasks)
+    tasks.each do |entry|
+      Iodine.run_every((entry[:interval] * 1000).to_i) do
+        entry[:task].enqueue
+      end
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


Abishekcs:
I have kept the design proposal same as suggested by @ShashantNagpure and @pratyush07-hub

lib/rage/deferred/scheduler.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class Rage::Deferred::Scheduler
+  LOCK_PATH = "/tmp/rage_deferred_scheduler.lock"
+
+  def self.start(tasks)
+    return if tasks.empty?
+
+    elect_leader { register_timers(tasks) }
```
rsamoilov: There's already an existing `Rage::Internal.pick_a_worker` method - you can use it instead.

lib/rage/configuration.rb
```diff
@@ -707,6 +707,22 @@ class Deferred
...
+    end
+
+    # Evaluates the scheduling DSL block, making `every` available as a method
+    def schedule(&block)
+      instance_eval(&block)
+    end
+
+    # Registers a task to run on a fixed interval (in seconds)
+    def every(interval, task:)
+      @scheduled_tasks << { interval:, task: }
```
rsamoilov: Do you think it makes sense to validate that `task` is a class that includes `Rage::Deferred::Task`?

lib/rage/configuration.rb
```diff
@@ -707,6 +707,22 @@ class Deferred
     # @private
     def initialize
       @configured = false
+      @scheduled_tasks = []
+    end
+
+    # Evaluates the scheduling DSL block, making `every` available as a method
+    def schedule(&block)
+      instance_eval(&block)
```
rsamoilov: This won't work with user-level constants - they are loaded after the framework is configured. So if a background task is defined in `app/tasks`, it won't be available during the configuration phase.

To fix this you'll need to update `schedule` to only store the block, postponing the execution of this block to the time when the scheduler starts. Also, keep in mind there could be multiple `schedule` calls.

rsamoilov (commented):
Hi @Abishekcs ,

Left several comments. Apart from that, this looks great!

lib/rage/configuration.rb
Abishekcs: That makes sense, since it will prevent `newbie users` like me from passing anything other than `Rage::Deferred::Task` when using Rails with Rage.

lib/rage/deferred/scheduler.rb
Abishekcs: Okay.

lib/rage/configuration.rb
Abishekcs: Thanks for the review. I will make the required changes.

Abishekcs:
Hi @rsamoilov, made the required changes based upon the comments you left, here is a quick summary:

- I had to modify `pick_a_worker` because the old version always used `Iodine.on_state(:on_start)` to register the lock attempt. But `deferred.rb` calls `__initialize` immediately when `Iodine.running?` is true or it will call it's own `Iodine.on_state(:on_start)` meaning `pick_a_worker` would be called while Iodine is already running. In that case `on_state(:on_start)` never fires and the block never executes. The fix adds an `Iodine.running?` check inside `pick_a_worker` itself so it calls the block immediately when Iodine is already running, and falls back to `on_state(:on_start)` otherwise.

- Changes are made to `schedule method` so it could only stores the scheduling block for later execution and added a `class ScheduleDSL under class deferred in confiuration.rb` to keep the DSL method `every` isolated in its own class rather than polluting Deferred with methods that don't belong there.

- Updated the Screenrecording and PR details.

- **AI usage**
  - `spec/internal_spec.rb` and `spec/deferred/scheduler_spec.rb` for suggestion test cases.

rsamoilov (approved):
Great change! Thank you! 🚀

PR #286, opened 2026-05-07
Added a second argument for head method

numice:
Added a second argument for `head` method in `controller/api.rb`

### DIFF
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -551,12 +551,19 @@ class RageController::API
   # Send a response with no body.
   #
   # @param status [Integer, Symbol] set a response status
+  # @param options [Symbol] pass extra options to the response headers
   # @example
   #   head :unauthorized
   # @example
   #   head 429
   def head(status, options = nil)
     @__rendered = true
+
+    if options
+      options.each do |key, value|
+        @__headers[key] = value.to_s
+      end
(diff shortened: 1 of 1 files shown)


numice:
I'm not exactly sure if that's right and if some tests are needed. On the first look, I thought that was going to involve many changes. Then I got busy for awhile. Now I took a look again and that's how I understand the task is. Well, it's just one change. But I don't know if anything else is supposed to be added.

rsamoilov:
Hi @numice 

Happy to see you back!

This is a good start. The next step would be to implement a code that will merge `options` (which is essentially a hash holding response headers the user wants to set) into the response headers that the framework will pass back to Rack.

numice:
@rsamoilov Got it. My bad I didn't look at the source code in Rails close enough to get this.

numice:
@rsamoilov Hi. Sorry for the delay. I've updated it based on my understanding. Not totally sure if this is it.

rsamoilov (commented):
That's it!

How are you feeling about covering this with tests? I'll be happy to merge once there's test coverage for the change.

Let me know if you need any help.

numice:
Thank you for taking a look at it. I was actually thinking about adding a test but not entirely sure how to go with it. I will take a look at the existing tests and then try to come up with something.

Issue #287, opened 2026-05-10
Blueprinter parser scaffold and class detection


rsamoilov:
Create `lib/rage/open_api/parsers/blueprinter.rb`, wire it into the serializer detection logic alongside the Alba parser. Activate when a referenced class inherits from `Blueprinter::Base`. Implement the parser entry point, file-loading strategy, and a minimal smoke test.

```ruby
class UserBlueprint < Blueprinter::Base
  # -> parser activates, returns {} for now
end
```

rsamoilov:
Implemented in #298

PR #298, opened 2026-05-13
[OpenAPI]: Add Blueprinter parser scaffold and class detection

Abishekcs:
### What this PR does? #287

Adds the Blueprinter parser scaffold and wires it into the existing  serializer detection logic alongside Alba.

The parser activates when it sees a class inheriting from `Blueprinter::Base` and returns `{ "type" => "object" }` for now full field extraction will come in follow-up PRs.


rsamoilov (approved):
👍

PR #299, opened 2026-05-16
[OpenAPI]: Extend @response / @request tag syntax to accept serializer options

Abishekcs:
### What this PR does

Extends the tag parser to recognize serializer options like **(Few Examples)** :
a) `UserBlueprint`
b) `UserBlueprint(view: :extended)`
c)  `UserBlueprint(view: :extended, root: :user)`
d)  `Array<UserBlueprint>`
e)  `Array<UserBlueprint(view: :extended, root: :user)`
... .

As, recmoneded in the issue description changes have been made in a way so Alba and any future serializer can use the same syntax.

### References

- Blueprinter [Usage](https://github.com/procore-oss/blueprinter#usage) Documentation.

### AI usage

- Used AI for generating the the `test cases`. I also went through all the test cases that was generated by the AI just to be sure it's okay.

### DIFF
--- a/lib/rage/openapi/openapi.rb
+++ b/lib/rage/openapi/openapi.rb
@@ -128,14 +128,15 @@ module Rage::OpenAPI
   end
 
   # @private
+  # @return [Array<Boolean, String, Hash>] a tuple of (is_collection, serializer, args)
   def self.__parse_serializer_args(str)
     is_collection, inner = __try_parse_collection(str)
 
     if is_collection
       # discard is_collection since we already know this is a collection from the outer call
-      _, clean_inner, options = __parse_serializer_args(inner)
-      if options.any?
-        [is_collection, clean_inner, options]
+      _, clean_inner, args = __parse_serializer_args(inner)
+      if args.any?
+        [is_collection, clean_inner, args]
       else
         [is_collection, clean_inner, {}]
       end
--- a/lib/rage/openapi/openapi.rb
+++ b/lib/rage/openapi/openapi.rb
@@ -128,6 +128,36 @@ module Rage::OpenAPI
   end
 
   # @private
+  def self.__parse_serializer_args(str)
+    is_collection, inner = __try_parse_collection(str)
+
+    if is_collection
+      # discard is_collection since we already know this is a collection from the outer call
+      _, clean_inner, options = __parse_serializer_args(inner)
+      if options.any?
+        [is_collection, clean_inner, options]
+      else
+        [is_collection, clean_inner, {}]
+      end
+    elsif str =~ /^([\w:]+)\(([^)]+)\)$/
+      [is_collection, $1, __parse_keywords($2)]
+    else
+      [is_collection, str, {}]
+    end
+  end
+
+  # @private
+  def self.__parse_keywords(str)
+    return {} if str.nil? || str.empty?
+
+    str.split(",").each_with_object({}) do |part, hash|
+      option = YAML.load(part)
+      return nil unless option.is_a?(Hash)
+      hash.merge!(option.transform_keys!(&:to_sym))
+    end
+  end
+
+  # @private
   def self.__module_parent(klass)
     klass.name =~ /::[^:]+\z/ ? Object.const_get($`) : Object
   rescue NameError
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ### Added
 
 - Implement `FiberScheduler#fiber_interrupt` (#283).
-- Add Blueprinter parser scaffold and class detection (#287)
+- [OpenAPI] Add Blueprinter parser scaffold and class detection (#287)
+- [OpenAPI] Extend @response / @request tag syntax to accept serializer options (#299)
 
 ## [1.24.0] - 2026-05-12
 
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -7,7 +7,7 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
   end
 
   def known_definition?(str)
-    _, str = Rage::OpenAPI.__try_parse_collection(str)
+    _, str, _ = Rage::OpenAPI.__parse_serializer_args(str)
     defined?(Blueprinter::Base) && @namespace.const_get(str).ancestors.include?(Blueprinter::Base)
   rescue NameError
     false
--- a/spec/openapi/openapi_spec.rb
+++ b/spec/openapi/openapi_spec.rb
@@ -179,6 +179,11 @@ RSpec.describe Rage::OpenAPI do
       let(:str) { "view: :extended, root: :user" }
       it { is_expected.to eq({ view: :extended, root: :user }) }
     end
+
+    context "with an invalid option (not a key value pair)" do
+      let(:str) { "extended" }
+      it { is_expected.to be_nil }
+    end
   end
 
   describe ".__module_parent" do
--- a/spec/openapi/openapi_spec.rb
+++ b/spec/openapi/openapi_spec.rb
@@ -73,6 +73,114 @@ RSpec.describe Rage::OpenAPI do
     end
   end
 
+  describe ".__parse_serializer_options" do
+    subject { described_class.__parse_serializer_args(str) }
+
+    context "with a plain class name" do
+      let(:str) { "UserBlueprint" }
+      it { is_expected.to eq([false, "UserBlueprint", {}]) }
+    end
+
+    context "with a view option" do
+      let(:str) { "UserBlueprint(view: :extended)" }
+      it { is_expected.to eq([false, "UserBlueprint", { view: :extended }]) }
+    end
+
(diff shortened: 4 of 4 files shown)


lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,48 @@ def self.__try_parse_collection(str)
...
+        hash[key.to_sym] = value[1..].to_sym
+      elsif value.start_with?('"') && value.end_with?('"')
+        hash[key.to_sym] = value[1..-2]
+      elsif value == "true"
+        hash[key.to_sym] = true
+      elsif value == "false"
+        hash[key.to_sym] = false
+      else
+        hash[key.to_sym] = value
+      end
```
rsamoilov: Can we substitute this logic with YAML parsing? For example:


```suggestion
      option = YAML.load(part)
      return nil unless option.is_a?(Hash) # TODO: handle the nil response in `OpenAPI::Parser`

      hash.merge!(option.transform_keys!(&:to_sym))
```

lib/rage/openapi/parser.rb
```diff
@@ -210,10 +210,13 @@ def parse_response_tag(expression, node, comment)
     elsif response_data.nil?
       node.responses[status] = nil
     else
+      response_data, serializer_options = Rage::OpenAPI.__parse_serializer_options(response_data)
```
rsamoilov: If you check the existing parsers, you'll see that each of them internally calls the similar `__try_parse_collection` call.

This creates a fair amount of duplication, but this is by design - parsers are intended to be independent systems that receive raw response tags and decide what to do with them.

That being said, let's move this call inside the Blueprinter parser. It's currently the only one using these options, which will also allow us to leave out the `**` signature updates for parsers that don't support options. This will also allow the parsers to correctly issue a warning if someone passes options to the parser that doesn't support them.

lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,48 @@ def self.__try_parse_collection(str)
     end
   end
 
+  # @private
+  def self.__parse_serializer_options(str)
```
rsamoilov: What about `args` instead of `options`?

```suggestion
  def self.__parse_serializer_args(str)
```

lib/rage/openapi/openapi.rb
Abishekcs: Okay, I didn't knew about `YAML.load ...`  😅 make things much simpler. Thanks.

lib/rage/openapi/parser.rb
Abishekcs: Got it.

Abishekcs:
Hi @rsamoilov can you review the changes again whenever you're free. Thank you! Here's a quick summary:

- Moved the `.__parse_serializer_args` call (previously `.__parse_serializer_options`) from `parser.rb` into `blueprinter.rb`
- Removed `**` from methods where it was added
- `.__parse_serializer_args` (previously `.__parse_serializer_options`) used to return `[str, options]`, now changed to return `[is_collection, klass_str, options]` where `klass_str` is the clean serializer class name stripped of any collection syntax or options
- Updated test case.

lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,36 @@ def self.__try_parse_collection(str)
...
 
+  # @private
+  def self.__parse_serializer_args(str)
+    is_collection, inner = __try_parse_collection(str)
+
+    if is_collection
+      # discard is_collection since we already know this is a collection from the outer call
+      _, clean_inner, options = __parse_serializer_args(inner)
+      if options.any?
+        [is_collection, clean_inner, options]
```
rsamoilov:
```suggestion
      _, clean_inner, args = __parse_serializer_args(inner)
      if args.any?
        [is_collection, clean_inner, args]
```

lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,36 @@ def self.__try_parse_collection(str)
     end
   end
 
+  # @private
+  def self.__parse_serializer_args(str)
```
rsamoilov, line 132:
```suggestion
  # @private
  # @return [Array<Boolean, String, Hash>] a tuple of (is_collection, serializer, args)
  def self.__parse_serializer_args(str)
```

lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,36 @@ def self.__try_parse_collection(str)
...
+    is_collection, inner = __try_parse_collection(str)
+
+    if is_collection
+      # discard is_collection since we already know this is a collection from the outer call
+      _, clean_inner, options = __parse_serializer_args(inner)
+      if options.any?
+        [is_collection, clean_inner, options]
+      else
+        [is_collection, clean_inner, {}]
```
rsamoilov, line 141: This branch is not covered with tests.

lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,36 @@ def self.__try_parse_collection(str)
...
+    elsif str =~ /^([\w:]+)\(([^)]+)\)$/
+      [is_collection, $1, __parse_keywords($2)]
+    else
+      [is_collection, str, {}]
+    end
+  end
+
+  # @private
+  def self.__parse_keywords(str)
+    return {} if str.nil? || str.empty?
```
rsamoilov: The `str.empty?` clause is not covered with tests.

lib/rage/openapi/openapi.rb
```diff
@@ -127,6 +127,36 @@ def self.__try_parse_collection(str)
...
+
+    str.split(",").each_with_object({}) do |part, hash|
+      option = YAML.load(part)
+      return nil unless option.is_a?(Hash)
```
rsamoilov, line 156: This condition is not covered with tests.

rsamoilov (commented):
Hey @Abishekcs,

This looks really nice.

I left a couple of minor suggestions - one to add a YARD comment and another to update the name of the variable.

There're also multiple statements not covered with tests. Let's either remove them if the situations they guard against can't happen or add tests to cover them.

Abishekcs:
>There're also multiple statements not covered with tests. Let's either remove them if the situations they guard against can't happen or add tests to cover them.

I have to check this one to be sure. Thanks for the review.

lib/rage/openapi/openapi.rb
Abishekcs: It's covered by this two test [(Line 94)](https://github.com/rage-rb/rage/pull/299/changes#diff-822e3625d8b3af53d38a1bc59f4649969ee95d37b4abfdfcaf999668bbb94f14R94) and [(Line 99)](https://github.com/rage-rb/rage/pull/299/changes#diff-822e3625d8b3af53d38a1bc59f4649969ee95d37b4abfdfcaf999668bbb94f14R99) cases:

```ruby
context "with a collection using square brackets without options" do
  let(:str) { "[UserBlueprint]" }
  it { is_expected.to eq([true, "UserBlueprint", {}]) }
end

context "with a collection using Array syntax without options" do
  let(:str) { "Array<UserBlueprint>" }
  it { is_expected.to eq([true, "UserBlueprint", {}]) }
end
```
- https://rage-rb.dev/docs/openapi#collections

lib/rage/openapi/openapi.rb
Abishekcs: This is the test case [(Line 148)](https://github.com/rage-rb/rage/pull/299/changes#diff-822e3625d8b3af53d38a1bc59f4649969ee95d37b4abfdfcaf999668bbb94f14R148):

```ruby
context "with empty string" do
  let(:str) { "" }
  it { is_expected.to eq({}) }
end
```

While poking around I noticed removing `str.empty?` clause didn't break the test `context "with empty string" ` because `"".split(",")` returns an empty array `[]`, so `each_with_object({})` just returns `{}` naturally without the block ever running.

image



So `str.empty?` is technically redundant, but I think  keeping it will be nice since it's not immediately obvious that `"".split(",").each_with_object({})` returns `{}` , the explicit check makes the intent clearer for anyone reading the code later.

Abishekcs: Does this make sense😅 ? or should i remove `str.empty?`

lib/rage/openapi/openapi.rb
Abishekcs: Done

lib/rage/openapi/openapi.rb
Abishekcs: Done

lib/rage/openapi/openapi.rb
Abishekcs: Added a new test case for this [(Line 183)](https://github.com/rage-rb/rage/pull/299/changes#diff-822e3625d8b3af53d38a1bc59f4649969ee95d37b4abfdfcaf999668bbb94f14R183)

```ruby
context "with an invalid option (not a key value pair)" do
  let(:str) { "extended" }
    it { is_expected.to be_nil }
  end
```

roman-softserve (approved):
Looks great!

rsamoilov (approved):

PR #305, opened 2026-06-01
[API] Reject malformed or empty HTTP token authorization headers

### LINKED ISSUE #304: [API] Reject malformed or empty HTTP token authorization headers
## Description:

Currently, `RageController::API#authenticate_with_http_token` accepts malformed `Authorization` headers as valid token input.

The helper only checks whether the header starts with `Bearer` or `Token`, then slices the string at fixed offsets. Because of that:

- `Authorization: Bearermy_token` is treated as a valid Bearer token header and yields a truncated token
- `Authorization: Tokenmy_token` is treated as a valid Token header and yields a truncated token
- `Authorization: Bearer ` and `Authorization: Token ` still invoke the authentication block with an empty token

The method documentation says it should return `nil` when no token is found. Malformed or empty token headers should therefore be rejected instead of being passed to the authentication block.

## Minimal Reproduction Example (MRE):

```ruby
controller = RageController::API.new({ "HTTP_AUTHORIZATION" => "Bearermy_token" }, nil)

result = controller.authenticate_with_http_token { |token| token }
# current result: "y_token"
```

Another example:

```ruby
controller = RageController::API.new({ "HTTP_AUTHORIZATION" => "Bearer token=" }, nil)

result = controller.authenticate_with_http_token { :authenticated }
# current result: :authenticated
```

## Current Behavior

Malformed or empty token headers are treated as valid input.

For example:

```http
Authorization: Bearermy_token
Authorization: Tokenmy_token
Authorization: Bearer
Authorization: Token
Authorization: Bearer 
Authorization: Token 
Authorization: Bearer token=
Authorization: Token token=""
```

The helper still yields to the authentication block for these inputs.

## Expected Behavior

These malformed or empty headers should return `nil` and should not yield to the authentication block.

These valid formats should keep working unchanged:

```http
Authorization: Bearer my_token
Authorization: Bearer token=my_token
Authorization: Token my_token
Authorization: Token token="my_token"
```

### The issue reproduction screenshot:

Image

RedZapdos123:
## Description:

Issue #304 reported that malformed or empty HTTP token `Authorization` headers could still be accepted by `RageController::API#authenticate_with_http_token` because the helper only checked the `Bearer` / `Token` prefix and then sliced the header at fixed offsets.

This PR updates the token authentication helper to require the expected scheme separator (`Bearer ` / `Token `) and to return `nil` when the extracted token is empty.

It also adds regression tests covering malformed headers without a separator and headers with empty token payloads, while preserving the existing valid Bearer and Token formats.

Closes #304.

## Checklist:

- [x] I have run linting checks in WSL using `rubocop --except Layout/EndOfLine lib/rage/controller/api.rb spec/controller/api/authenticate_spec.rb`.
- [x] I have run focused tests in WSL using `rspec spec/controller/api/authenticate_spec.rb`.
- [x] I have run broader API regression tests in WSL using `rspec spec/controller/api spec/rage/response_spec.rb`.

### Before the fix:
image

### After the fix:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 - [OpenAPI] Add Blueprinter parser scaffold and class detection (#287)
 - [OpenAPI] Extend @response / @request tag syntax to accept serializer options (#299)
 
+### Fixed
+
+- [API] Reject malformed or empty HTTP token authorization headers.
+
 ## [1.24.0] - 2026-05-12
 
 ### Added
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -558,13 +558,13 @@ class RageController::API
   def authenticate_with_http_token
     auth_header = @__env["HTTP_AUTHORIZATION"]
 
-    payload = if auth_header&.start_with?("Bearer")
+    payload = if auth_header&.start_with?("Bearer ")
       auth_header[7..]
-    elsif auth_header&.start_with?("Token")
+    elsif auth_header&.start_with?("Token ")
       auth_header[6..]
     end
 
-    return unless payload
+    return if payload.nil? || payload.empty?
 
     token = if payload.start_with?("token=")
       payload[6..]
@@ -574,6 +574,8 @@ class RageController::API
 
     token.delete_prefix!('"')
     token.delete_suffix!('"')
+
+    return if token.empty?
 
     yield token
   end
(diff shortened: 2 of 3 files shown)


rsamoilov (approved):
Hi @RedZapdos123 ,

This looks great. Thank you for the contribution! Hope to see more changes from you!

PR #307, opened 2026-06-01
[Router] Normalize exact host constraints when host includes a port

### LINKED ISSUE #306: [Router] Exact host constraints do not match when request host includes a port
### Description:

Exact host constraints currently fail when the request host includes a port.

The router derives the `host` constraint directly from `HTTP_HOST`, so a route constrained with `host: "google.com"` does not match a request with `HTTP_HOST = "google.com:3000"`.

```ruby
class PhotosController < RageController::API
  def index
    render plain: "ok"
  end
end

Rage.routes.draw do
  get "/photos", to: "photos#index", constraints: { host: "google.com" }
end
```

Request:

```text
GET /photos
Host: google.com:3000
```

### Current behavior

The route does not match, because the router compares the exact constraint value `"google.com"` against the raw `HTTP_HOST` value `"google.com:3000"`.

### Expected behavior

The route should match.

The port should not affect exact host constraint matching. Rage already exposes the normalized host separately in `Rage::Request#host`, so host constraints should follow the same behavior.

### The issue reproduction:

Image

RedZapdos123:
## Description:

This fixes exact host constraint matching when the request host includes a port.

The router currently derives the `host` constraint directly from `HTTP_HOST`, so a route constrained with `host: "google.com"` does not match requests like `Host: google.com:3000`.

This PR strips the port from `HTTP_HOST` before host constraint matching, so exact host constraints compare against the normalized host value.

It also adds a regression spec covering requests where the host includes a port.

Closes #306.

## Checklist:

- [x] I have run linting checks in WSL using `rubocop --except Layout/EndOfLine lib/rage/router/constrainer.rb spec/router/constraints_spec.rb`.
- [x] I have run focused tests in WSL using `rspec spec/router/constraints_spec.rb`.
- [x] I have run broader related tests in WSL using `rspec spec/router`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - [API] Reject malformed or empty HTTP token authorization headers.
+- [Router] Strip the port from `HTTP_HOST` before matching exact host constraints.
 
 ## [1.24.0] - 2026-05-12
 
--- a/lib/rage/router/constrainer.rb
+++ b/lib/rage/router/constrainer.rb
@@ -69,7 +69,7 @@ class Rage::Router::Constrainer
       # Optimization: inline the derivation for the common built in constraints
       if !strategy.custom?
         if key == :host
-          lines << "   host: env['HTTP_HOST'.freeze],"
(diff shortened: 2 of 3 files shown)


rsamoilov (approved):
Looks great!

PR #309, opened 2026-06-02
[Cookies] Preserve array domain matching when request host includes a port

### LINKED ISSUE #308: [Cookies] Array domain matching fails when request host includes a port
### Description:

Cookie domain matching for array-based domains currently fails when the request host includes a port.

`Rage::Cookies#[]=` compares the configured domain list against the raw `HTTP_HOST`, so a cookie setup like `domain: %w(api.test.com cookie.test.com)` does not match a request with `HTTP_HOST = "cookie.test.com:3000"`.

```ruby
class CookiesController < RageController::API
  def create
    cookies[:user_id] = {
      domain: %w(api.test.com cookie.test.com),
      value: 120
    }

    render plain: "ok"
  end
end

Rage.routes.draw do
  get "/cookies", to: "cookies#create"
end
```

Request:

```text
GET /cookies
Host: cookie.test.com:3000
```

### Current behavior:

The cookie is set, but the `domain` attribute is omitted:

```text
Set-Cookie: user_id=120
```

For comparison, the same setup works correctly when the host does not include a port:

```text
Host: cookie.test.com
Set-Cookie: user_id=120; domain=cookie.test.com
```

### Expected behavior:

The configured domain should still match, and the response should include the domain attribute:

```text
Set-Cookie: user_id=120; domain=cookie.test.com
```

The port should not affect array based cookie domain matching.

### The issue reproduction:

Image

RedZapdos123:
## Description:

Issue #308 reported that array-based cookie domain matching could fail when the request host included a port.

This PR updates `Rage::Cookies` to strip the port from `HTTP_HOST` before resolving configured cookie domains.

It also adds a regression test that covers array-based cookie domain matching when the request host includes a port.

Closes #308.

##Checklist:

- [x] I have run linting checks in WSL using `rubocop --except Layout/EndOfLine lib/rage/cookies.rb spec/controller/api/cookies_spec.rb`.
- [x] I have run focused tests in WSL using `rspec spec/controller/api/cookies_spec.rb`.
- [x] I have run broader related tests in WSL using `rspec spec/controller/api`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - [API] Reject malformed or empty HTTP token authorization headers.
+- [Cookies] Strip the port from `HTTP_HOST` before matching configured cookie domains.
 - [Router] Strip the port from `HTTP_HOST` before matching exact host constraints.
 
 ## [1.24.0] - 2026-05-12
--- a/lib/rage/cookies.rb
+++ b/lib/rage/cookies.rb
@@ -201,7 +201,7 @@ class Rage::Cookies
     end
 
     if (domain = value[:domain])
-      host = @env["HTTP_HOST"]
+      host = @env["HTTP_HOST"]&.sub(/:\d+\z/, "")
 
       processed_domain = if domain.is_a?(String)
         domain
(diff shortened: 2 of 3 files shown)


rsamoilov (approved):
Looks great! Thank you so much!

PR #312, opened 2026-06-03
[Request] Treat IPv6 literals as non-domain hosts

### LINKED ISSUE #311: [Request] Treat IPv6 literals as non-domain hosts
### Description:

`Rage::Request#domain` currently treats IPv6 host literals as domain names instead of IP literals.

With `HTTP_HOST = "[::1]:3000"`, `request.domain` returns `"[::1]"`. In the fallback path where `HTTP_HOST` is missing and `SERVER_NAME = "::1"`, `request.domain` can also raise.

```ruby
env = {
  "rack.upgrade?" => nil,
  "rack.upgrade" => nil,
  "rack.version" => [1, 3],
  "rage.request_id" => "x",
  "SCRIPT_NAME" => "",
  "rack.url_scheme" => "http",
  "HTTP_VERSION" => "HTTP/1.1",
  "PATH_INFO" => "/users",
  "QUERY_STRING" => "",
  "REMOTE_ADDR" => "::1",
  "REQUEST_METHOD" => "GET",
  "SERVER_PROTOCOL" => "HTTP/1.1",
  "IODINE_REQUEST_ID" => "x",
  "IODINE_HAS_BODY" => false,
  "SERVER_PORT" => "3000",
  "HTTP_HOST" => "[::1]:3000"
}

request = Rage::Request.new(env)

request.host
# => "[::1]"

request.domain
# => "[::1]"

fallback_env = env.dup
fallback_env.delete("HTTP_HOST")
fallback_env["SERVER_NAME"] = "::1"

Rage::Request.new(fallback_env).domain
# => raises NoMethodError
```

### Current behavior:

- `request.domain` returns `"[::1]"` for IPv6 host literals.
- `request.domain(0)` also returns `"[::1]"`.
- In the fallback path with bare IPv6 `SERVER_NAME`, `request.domain` raises `NoMethodError`.
- IPv4 literal hosts already return `nil`, so the behavior is inconsistent for IP hosts.

### Expected behavior:

- IPv6 literals should be treated like IPv4 literals.
- `request.domain` should return `nil` for IPv6 host literals.
- `request.domain(0)` should also return `nil` for IPv6 host literals.
- `request.domain` should not raise when the request falls back to an IPv6 `SERVER_NAME`.

### The issue repoduction:

Image

RedZapdos123:
## Description:

Issue #311 reported that `Rage::Request#domain` could treat IPv6 literals as domain names and could raise when the request fell back to a bare IPv6 `SERVER_NAME`.

This PR updates `Rage::Request` to detect IP literals using `IPAddr`, so IPv6 hosts are treated the same way as IPv4 hosts during domain extraction.

It also adds regression tests that cover IPv6 `HTTP_HOST` handling and the fallback path where `HTTP_HOST` is missing.

Closes #311.

## Checklist:

- [x] I have run linting checks in WSL using `rubocop --except Layout/EndOfLine lib/rage/request.rb spec/rage/request_spec.rb`.
- [x] I have run focused tests in WSL using `rspec spec/rage/request_spec.rb`.
- [x] I have run broader related tests in WSL using `rspec spec/rage`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - [API] Reject malformed or empty HTTP token authorization headers.
 - [Cookies] Strip the port from `HTTP_HOST` before matching configured cookie domains.
 - [Router] Strip the port from `HTTP_HOST` before matching exact host constraints.
+- [Request] Treat IPv6 literals as non-domain hosts.
 
 ## [1.24.0] - 2026-05-12
 
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "ipaddr"
 require "time"
 
 class Rage::Request
@@ -245,7 +246,14 @@ class Rage::Request
   end
 
   def named_host?(host)
-    !IP_HOST_REGEXP.match?(host)
+    host && !ip_host?(host)
+  end
+
+  def ip_host?(host)
+    IP_HOST_REGEXP.match?(host) || IPAddr.new(host.delete_prefix("[").delete_suffix("]"))
+    true
+  rescue IPAddr::InvalidAddressError
+    false
   end
 
   def if_none_match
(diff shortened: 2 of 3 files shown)


rsamoilov (approved):
Nice! 🚀

PR #314, opened 2026-06-03
[API] Reject invalid status symbols in head and render

### LINKED ISSUE #313: [API] Reject invalid status symbols in head and render
### Description:

`head` and `render` currently accept invalid status symbols and silently set the response status to `nil`.

Both methods convert symbol statuses through `Rack::Utils::SYMBOL_TO_STATUS_CODE`, but they do not validate that the symbol exists in the status table before assigning it.

```ruby
controller = RageController::API.new(nil, nil)

controller.head(:not_a_status)
controller.response.status
# => nil

controller = RageController::API.new(nil, nil)

controller.render plain: "ok", status: :not_a_status
controller.response.status
# => nil
```

### Current behavior:

- `head(:not_a_status)` does not raise.
- `render plain: "ok", status: :not_a_status` does not raise.
- in both cases, the response status becomes `nil`

Observed output on latest `main`:

```text
{method: "head", status: nil, body: "\"\""}
{method: "render", status: nil, body: "\"ok\""}
```

### Expected behavior:

Invalid status symbols should not be accepted silently.

### The issue reproduction:

Image

RedZapdos123:
## Description:

Issue #313 reported that `head` and `render` could accept invalid status symbols and silently set the response status to `nil`.

This PR updates `RageController::API` to validate symbol statuses through `Rack::Utils.status_code` before mutating controller state.

It also adds regression tests that cover invalid status symbols in both `head` and `render`, and verifies that the response status remains unchanged when validation fails.

Closes #313.

## Checklist:

- [x] I have run linting checks in WSL using `rubocop --except Layout/EndOfLine lib/rage/controller/api.rb spec/controller/api/render_spec.rb spec/rage/response_spec.rb`.
- [x] I have run focused tests in WSL using `rspec spec/controller/api/render_spec.rb spec/rage/response_spec.rb`.
- [x] I have run broader related tests in WSL using `rspec spec/controller/api spec/rage/response_spec.rb`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -485,7 +485,7 @@ class RageController::API
   # @note `render` doesn't terminate execution of the action, so if you want to exit an action after rendering, you need to do something like `render(...) and return`.
   def render(json: nil, plain: nil, sse: nil, status: nil)
     raise "Render was called multiple times in this action." if @__rendered
-    status = normalize_status(status) if status
+    status = ::Rack::Utils.status_code(status) if status
     @__rendered = true
 
     if json || plain
@@ -527,7 +527,7 @@ class RageController::API
   # @example
   #   head 429
   def head(status)
-    status = normalize_status(status)
+    status = ::Rack::Utils.status_code(status)
     @__rendered = true
     @__status = status
   end
@@ -540,11 +540,6 @@ class RageController::API
   def headers
     @__headers
   end
-
-  def normalize_status(status)
-    status.is_a?(Symbol) ? ::Rack::Utils.status_code(status) : status
-  end
-  private :normalize_status
 
   # Authenticate using an HTTP Bearer token. Returns the value of the block if a token is found. Returns `nil` if no token is found.
   #
--- a/lib/rage/controller/api.rb
+++ b/lib/rage/controller/api.rb
@@ -485,7 +485,7 @@ class RageController::API
   # @note `render` doesn't terminate execution of the action, so if you want to exit an action after rendering, you need to do something like `render(...) and return`.
   def render(json: nil, plain: nil, sse: nil, status: nil)
     raise "Render was called multiple times in this action." if @__rendered
-    status = ::Rack::Utils.status_code(status) if status
+    normalized_status = status.nil? ? nil : ::Rack::Utils.status_code(status)
     @__rendered = true
 
     if json || plain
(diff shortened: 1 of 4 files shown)


lib/rage/controller/api.rb
```diff
@@ -500,11 +501,7 @@ def render(json: nil, plain: nil, sse: nil, status: nil)
     end
 
     if status
-      @__status = if status.is_a?(Symbol)
-        ::Rack::Utils::SYMBOL_TO_STATUS_CODE[status]
-      else
-        status
-      end
+      @__status = status
```
rsamoilov:
```suggestion
      @__status = ::Rack::Utils.status_code(status)
```

rsamoilov (commented):
`::Rack::Utils.status_code` is already checking if status is a `Symbol`, which means you can simplify the code further and remove `normalize_status` completely.

lib/rage/controller/api.rb
RedZapdos123: Thank you for the suggestion. `Rack::Utils.status_code` already handles symbol validation and integer coercion, so removed the extra helper and keep the validation before controller state mutation.

I ran the focused specs, the broader controller/response suite, and RuboCop again. The invalid status fix still works as needed to.

rsamoilov: You’ve probably forgotten to push - I don’t see the changes.

RedZapdos123: Yes, I did forget. Pushed the changes in a commit now.

rsamoilov (commented):
This looks good, but is there any specific reason you chose to reassign the `status` variable instead assigning `@__status` directly?

Basically, that's what the code is currently doing:

```ruby
status = ::Rack::Utils.status_code(status)
@__status = status
```

Plus, there're two `if status` checks in `render` now. Can we do this?

```ruby
@__status = ::Rack::Utils.status_code(status)
```

RedZapdos123:
> This looks good, but is there any specific reason you chose to reassign the `status` variable instead assigning `@__status` directly?
> 
> Basically, that's what the code is currently doing:
> 
> ```ruby
> status = ::Rack::Utils.status_code(status)
> @__status = status
> ```
> 
> Plus, there're two `if status` checks in `render` now. Can we do this?
> 
> ```ruby
> @__status = ::Rack::Utils.status_code(status)
> ```

Thank you for the suggestions.

I simplified `head` to assign `@__status` directly through `Rack::Utils.status_code`.

For `render`, I tried the same shape, but it changed the failure behavior for invalid status symbols.  The response body was already written, and the status had become 200 before the exception was raised.

To keep the original fix, I am still validating the status before any controller state is mutated in `render` using a prevalidated local value there.

rsamoilov (approved):

rsamoilov:
> Thank you for the suggestions.
> 
> I simplified `head` to assign `@__status` directly through `Rack::Utils.status_code`.
> 
> For `render`, I tried the same shape, but it changed the failure behavior for invalid status symbols. The response body was already written, and the status had become 200 before the exception was raised.
> 
> To keep the original fix, I am still validating the status before any controller state is mutated in `render` using a prevalidated local value there.

Makes sense. Thank you!

PR #319, opened 2026-06-07
[Cookies] Use request host fallback when resolving cookie domains

### LINKED ISSUE #318: [Cookies] Use request host fallback when resolving cookie domains
### Description:

Cookie domain resolution currently depends on `HTTP_HOST` only.

When `HTTP_HOST` is missing and the request falls back to `SERVER_NAME`, `Rage::Cookies#[]=` does not resolve domains correctly:

- array-based domain matching silently skips the domain attribute
- `domain: :all` can raise because `DomainName(nil)` is called

This is inconsistent with `Rage::Request#host`, which already falls back to `SERVER_NAME`.

```ruby
require "domain_name"
require "rage/all"

headers = {}
env = {
  "SERVER_NAME" => "cookie.test.com",
  "SERVER_PORT" => "3000"
}

request = Rage::Request.new(env)
request.host
# => "cookie.test.com"

cookies = Rage::Cookies.new(env, headers)
cookies[:user_id] = {
  value: 120,
  domain: %w(api.test.com cookie.test.com)
}

headers["set-cookie"]
# => "user_id=120"

headers = {}
cookies = Rage::Cookies.new(env, headers)
cookies[:user_id] = {
  value: 120,
  domain: :all
}

# => raises TypeError: NilClass is not a String
```

### The issue reproduction:

Image

### Current behavior:

- With `SERVER_NAME = "cookie.test.com"` and no `HTTP_HOST`, array-based domain matching does not add the domain attribute:

```text
Set-Cookie: user_id=120
```

- With the same env and `domain: :all`, cookie assignment raises:

```text
TypeError: NilClass is not a String
```

### Expected behavior:

- Cookie domain resolution should use the same host fallback semantics as request host handling.
- When `SERVER_NAME = "cookie.test.com"`, this should work:

```text
Set-Cookie: user_id=120; domain=cookie.test.com
```

- `domain: :all` should not raise just because `HTTP_HOST` is absent.

RedZapdos123:
## Description:

Issue #318 reported that cookie domain resolution could depend on `HTTP_HOST` only.

This PR updates `Rage::Cookies` to derive the request host through `Rack::Request`, so cookie domain resolution follows the same host fallback semantics as request host handling when `HTTP_HOST` is missing and `SERVER_NAME` is present.

It also adds regression tests that cover both array-based domains and `domain: :all` when the request uses `SERVER_NAME` fallback.

Closes #318.

## Checklist:

- [x] I have run linting checks in WSL using `rubocop --except Layout/EndOfLine lib/rage/cookies.rb spec/controller/api/cookies_spec.rb`.
- [x] I have run focused tests in WSL using `rspec spec/controller/api/cookies_spec.rb`.
- [x] I have run broader related tests in WSL using `rspec spec/controller/api`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Fixed
 
 - [Request] Treat IPv6 literals as non-domain hosts.
+- [Cookies] Use request host fallback when resolving cookie domains.
 
 ## [1.25.0] - 2026-06-03
 
--- a/lib/rage/cookies.rb
+++ b/lib/rage/cookies.rb
@@ -201,7 +201,7 @@ class Rage::Cookies
     end
 
     if (domain = value[:domain])
-      host = @env["HTTP_HOST"]&.sub(/:\d+\z/, "")
+      host = Rack::Request.new(@env).host
 
       processed_domain = if domain.is_a?(String)
         domain
--- a/spec/controller/api/cookies_spec.rb
+++ b/spec/controller/api/cookies_spec.rb
@@ -6,10 +6,11 @@ require "domain_name"
 RSpec.describe RageController::API do
   subject do
     cookie_header = cookies.map { |k, v| "#{k}=#{v}" }.join("; ")
-    described_class.new({ "HTTP_COOKIE" => cookie_header, "HTTP_HOST" => request_host }, nil)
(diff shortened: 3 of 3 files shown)


rsamoilov (approved):
Thank you! 🚀

PR #321, opened 2026-06-08
[Router] Fall back to SERVER_NAME for exact host constraints

### LINKED ISSUE #320: [Router] Exact host constraints should fall back to SERVER_NAME
### Description:

Router exact host constraints currently derive the request host from `HTTP_HOST` only.

When `HTTP_HOST` is missing and the request falls back to `SERVER_NAME`, an exact host constraint like `host: "google.com"` does not match, even though the request host still resolves to `google.com`.

This is inconsistent with request host handling elsewhere in Rage, which already falls back to `SERVER_NAME`.

```ruby
require "rage/all"

router = Rage::Router::Backend.new
router.on("GET", "/photos", ->(_) { "photos" }, constraints: { host: "google.com" })

env = {
  "REQUEST_METHOD" => "GET",
  "PATH_INFO" => "/photos",
  "SERVER_NAME" => "google.com",
  "SERVER_PORT" => "3000",
  "rack.input" => StringIO.new
}

handler = router.lookup(env)
handler.nil?
# => true
```

### The issue reproduction:

Image

### Current behavior:

With `SERVER_NAME = "google.com"` and no `HTTP_HOST`, the exact host constraint does not match:

```text
handler.nil? # => true
```

### Expected behavior:

Exact host constraints should use the same host fallback semantics as request host handling.

With the same environment, the route should match:

```text
handler.nil? # => false
handler[:handler].call(env, handler[:params]) # => "photos"
```

RedZapdos123:
## Description:

Issue #320 reported that exact router host constraints only derived the request host from `HTTP_HOST`.

This PR updates the router constrainer to fall back to `SERVER_NAME` when `HTTP_HOST` is missing, while still stripping the port from `HTTP_HOST` when it is present.

It also adds a regression spec that covers exact host-constrained route lookup when the request host comes from `SERVER_NAME`.

Closes #320.

## Checklist:

- [x] I have run linting and formatting checks in WSL using `bundle exec rubocop --except Layout/EndOfLine lib/rage/router/constrainer.rb spec/router/constraints_spec.rb`.
- [x] I have run focused tests in WSL using `bundle exec rspec spec/router/constraints_spec.rb`.
- [x] I have run the broader router suite in WSL using `bundle exec rspec spec/router`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/lib/rage/router/constrainer.rb
+++ b/lib/rage/router/constrainer.rb
@@ -69,7 +69,7 @@ class Rage::Router::Constrainer
       # Optimization: inline the derivation for the common built in constraints
       if !strategy.custom?
         if key == :host
-          lines << "   host: env['HTTP_HOST'.freeze]&.sub(/:\\d+\\z/, ''.freeze),"
+          lines << "   host: (host = env['HTTP_HOST'.freeze]) ? host.sub(/:\\d+\\z/, ''.freeze) : env['SERVER_NAME'.freeze],"
         else
           raise ArgumentError, "unknown non-custom strategy for compiling constraint derivation function"
         end
--- a/lib/rage/router/constrainer.rb
+++ b/lib/rage/router/constrainer.rb
@@ -69,7 +69,7 @@ class Rage::Router::Constrainer
       # Optimization: inline the derivation for the common built in constraints
       if !strategy.custom?
         if key == :host
-          lines << "   host: (host = env['HTTP_HOST'.freeze]) ? host.sub(/:\\d+\\z/, ''.freeze) : env['SERVER_NAME'.freeze],"
+          lines << "   host: env['HTTP_HOST'.freeze]&.sub(/:\\d+\\z/, ''.freeze) || env['SERVER_NAME'.freeze],"
         else
           raise ArgumentError, "unknown non-custom strategy for compiling constraint derivation function"
         end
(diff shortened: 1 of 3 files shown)


lib/rage/router/constrainer.rb
```diff
@@ -69,7 +69,7 @@ def __build_derive_constraints
       # Optimization: inline the derivation for the common built in constraints
       if !strategy.custom?
         if key == :host
-          lines << "   host: env['HTTP_HOST'.freeze]&.sub(/:\\d+\\z/, ''.freeze),"
+          lines << "   host: (host = env['HTTP_HOST'.freeze]) ? host.sub(/:\\d+\\z/, ''.freeze) : env['SERVER_NAME'.freeze],"
```
rsamoilov: Would something like this work?

```suggestion
          lines << "   host: env['HTTP_HOST'.freeze]&.sub(/:\\d+\\z/, ''.freeze) || env['SERVER_NAME'.freeze],"
```

RedZapdos123: This does work.

The behavior stays the same for both the `SERVER_NAME` fallback path  and the current `HTTP_HOST` with port path, so I updated the PR to use the simpler expression.

The tests pass.

rsamoilov (approved):
Thank you!

PR #323, opened 2026-06-09
[OpenAPI]: Refactor: replace Prism AST parser with Blueprinter Reflection API

Abishekcs:
### What this PR does?

Switch the Blueprinter parser from Prism-based AST parsing to Blueprinter's built-in Reflection API. This removes the need for a Visitor pattern, temp file parsing, and manual AST traversal. The Reflection API provides direct access to fields, views, and associations making the parser significantly simpler.

### Example Code and Screenshot for basic Blueprinter fields: identifier, field, and fields

- `routes.rb` 

```ruby
Rage.routes.draw do
  root to: ->(env) { [200, {}, ["It works!"]] }

  get "/users/:id/blue", to: "users#show_blue"
end

```

- `products_controller.rb`

```ruby


class UsersController < RageController::API
  
  # @response UserBlueprint
  def show_blue
    user = [{ uuid: 1, id: 101, name: "Alice", email: "alice@example.com", age: 25, first_name: "Alice", last_name: "loom", abishek: 'abi' }]
    render json: UserBlueprint.render(user)
  end
end


```

- `user_blueprint.rb`

```ruby

class UserBlueprint < Blueprinter::Base
  identifier :uuid
  
  fields :id, :name, :email, :age
  
  field :email, name: :login   # -> "login": string

  field :abishek, name: :ad   # -> "login": string
  
  fields :first_name, :last_name

  field(:full_name) { |u| "#{u[:first_name]} #{u[:last_name]}" }
end


```

basic_fields

### Example Code and Screnshoot of Blueprinter Inheritance

- `routes.rb`

```ruby
Rage.routes.draw do
  root to: ->(env) { [200, {}, ["It works!"]] }

  get "/data_mining", to: "data_mining#show_data_mining"
end
```

- `data_mining_controller.rb`

```ruby

class DataMiningController < RageController::API

  # @response DataMining
  def show_data_mining
    data_mining_details = { uuid: 'Third Edition', email: 'MorganKaufmann@publishers.com', name: 'Han Kamber Pei', first_name: 'Morgan', last_name: 'Kaufmann', hello_world: 'Earth', subject: "DataMining"  }
    render json: DataMining.render(data_mining_details)
  end
end


```

- `data_mining_base.rb`

```ruby
  field :hello_world
  field :email, name: :something
  fields :subject
end

```

- `data_mining.rb`

```ruby
class DataMining < DataMiningBase
  identifier :uuid
  field "email", name: :login
  field "first_name"
  field :hello_world
end
```

- `Blueprinter Response for route /data_mining`

```
{
  "uuid": "Third Edition",
  "first_name": "Morgan",
  "hello_world": "Earth",
  "login": "MorganKaufmann@publishers.com",
  "something": "MorganKaufmann@publishers.com",
  "subject": "DataMining"
}
```

- `OpenAPI response output for # @response DataMining`

```ruby

{
  "uuid": "string",
  "first_name": "string",
  "hello_world": "string",
  "login": "string",
  "something": "string",
  "subject": "string"
}

```

blueprinter_inheritance


lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -14,107 +14,31 @@ def known_definition?(str)
...
-      @context.strings << node.unescaped
-    end
-
-    private
-
-    def with_context
-      @context = VisitorContext.new
-      yield
-      @context
+    view.instance_variable_get(:@view_collection).instance_variable_get(:@views)[view_name].instance_variable_get(:@fields).each_with_object({}) do |(_, field), hash|
```
Abishekcs: At first I was using something like this:


```ruby

    reflections = klass.reflections

    identifier_fields = {}
    default_fields = {}

    # identifier fields
    if (identifier_view = reflections[:identifier])
      identifier_view.fields.each do |_, field|
        identifier_fields[field.name.to_s] = { "type" => "string" }
      end
    end

    # default view fields
    if (default_view = reflections[:default])
      default_view.fields.each do |_, field|
        default_fields[field.name.to_s] = { "type" => "string" }
      end
    end

```

But `.fields` of Relection API seems to break for 

```ruby 
 fields "first_name", :last_name
```

Image


or for this too

```ruby
field :email, name: "login"
```

Image

Abishekcs:
@rsamoilov I have refactor Blueprinter parser using `Reflection API` just a quick summary:

- Added Gem `Blueprinter`
- Added a new mock class i.e `spec/support/contexts/mocked_blueprinter_classes.rb`. I am not sure if this was the right way or I should have made changes to the existing `mocked_classess.rb`.
- Existing test cases logic is same with just some minor changes.

Abishekcs:
Only works with 
- basic Blueprinter fields: identifier, field, and fields. 
- Blueprinter Inheritance

spec/support/contexts/mocked_blueprinter_classes.rb
```diff
@@ -0,0 +1,28 @@
...
+          class #{class_name} #{"< #{parent.name}" if parent != Object}
+            #{block.call}
+          end
+        RUBY
+      end
+      f.close
+    end
+
+    klass = Class.new(parent)
+    klass.class_eval(block.call) if block
```
Abishekcs: This is the only newly added line. Rest is same as `mocked_classes.rb`

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: > fields "first_name", :last_name

This seems to work fine for me with Blueprinter 1.2.1. But in any way, with this approach, we should only be relying on the official API. If some cases are not supported by the reflection API, it's safe to ignore them.

spec/support/contexts/mocked_blueprinter_classes.rb
rsamoilov: It should be safe to add this line to `mocked_classes.rb`. And you won't need a separate context then.

Abishekcs:
@rsamoilov PR is ready to be reviewed again. Thank you!

- This [test case](https://github.com/rage-rb/rage/pull/323/changes/87977ff9eadfc48a161447d8d3e9a7a6234566bc#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3L222) was fully removed 
- Switched back to using the [official API](https://github.com/rage-rb/rage/pull/323/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3L245)
- Removed `mocked_blueprinter_classes.rb` and added few new lines in `mocked_classes.rb`
  which is: 
  ```ruby 
    if block
      if defined?(Blueprinter::Base) && parent.ancestors.include?(Blueprinter::Base)
        klass.class_eval(block.call)
      else
        klass.class_eval(&block)
      end
    end
  ```

spec/support/contexts/mocked_classes.rb
```diff
@@ -25,6 +25,15 @@ class #{class_name} #{"< #{parent.name}" if parent != Object}
...
 
     klass = Class.new(parent, &block)
+
+    if block
+      if defined?(Blueprinter::Base) && parent.ancestors.include?(Blueprinter::Base)
+        klass.class_eval(block.call)
+      else
+        klass.class_eval(&block)
+      end
+    end
```
rsamoilov, line 35:
```suggestion
    if block
      if defined?(Blueprinter::Base) && parent.ancestors.include?(Blueprinter::Base)
        klass.class_eval(block.call)
      end
    end
```

The `else` branch doesn't seem to be used, so it'd be better to remove it in the next PR.

rsamoilov (approved):
🚀

PR #325, opened 2026-06-11
[API] Use weak comparison for If-None-Match validation

### LINKED ISSUE #324: [API] Use weak comparison for If-None-Match validation
### Description:

`Rage::Request#fresh?` currently compares `If-None-Match` values against the response ETag using exact string equality.

Rage emits weak ETags like `W/"..."`, but when the request sends the matching strong form like `"..."`, the request is treated as stale and the action still renders the body.

For `If-None-Match`, weak comparison should be used for normal GET and HEAD validation.

```ruby
require "digest"
require "rage/all"

class TestController < RageController::API
  def stale_etag_test
    return unless stale?(etag: "123")

    render plain: "body"
  end
end

handler = TestController.__register_action(:stale_etag_test)
env = {
  "HTTP_IF_NONE_MATCH" => "\"#{Digest::SHA1.hexdigest("123")}\""
}

TestController.new(env, {}).public_send(handler)
# => [200, {"content-type"=>"text/plain; charset=utf-8", "etag"=>"W/\"40bd001563085fc35165329ea1ff5c5ecbdbbeef\"", "last-modified"=>nil}, ["body"]]
```

### The issue reproduction:

Image

### Current behavior:

When the request sends a strong `If-None-Match` value that matches the response ETag value apart from the weak `W/` prefix, Rage renders the resource:

```text
[200, {"content-type"=>"text/plain; charset=utf-8", "etag"=>"W/\"40bd001563085fc35165329ea1ff5c5ecbdbbeef\"", "last-modified"=>nil}, ["body"]]
```

### Expected behavior:

For GET and HEAD validation, `If-None-Match` should use weak comparison.

With the same request, the matching strong form should be treated as not modified against the weak response ETag, and Rage should return:

```text
[304, {"etag"=>"W/\"40bd001563085fc35165329ea1ff5c5ecbdbbeef\"", ...}, []]
```

RedZapdos123:
## Description:

Issue #324 reported that `If-None-Match` validation used exact string comparison, so a strong request ETag like `"..."` did not match Rage's weak response ETag `W/"..."`.

This PR updates `Rage::Request#etag_matches?` to use weak comparison for `If-None-Match` validation, while keeping wildcard handling unchanged.

It also updates the conditional GET regression spec to cover the strong-request / weak-response match and adds a changelog entry.

Closes #324.

## Checklist:

- [x] I have run linting checks in WSL using `bundle exec rubocop --except Layout/EndOfLine lib/rage/request.rb spec/controller/api/conditional_get_spec.rb`.
- [x] I have run focused tests in WSL using `bundle exec rspec spec/controller/api/conditional_get_spec.rb`.
- [x] I have run a broader controller API regression pass in WSL using `bundle exec rspec spec/controller/api`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -271,8 +271,10 @@ class Rage::Request
 
     return true if requested_etags.empty?
     return false if response_etag.nil?
+    return true if requested_etags.include?("*")
 
-    requested_etags.include?(response_etag) || requested_etags.include?("*")
+    response_etag = weak_etag(response_etag)
+    requested_etags.any? { |requested_etag| weak_etag(requested_etag) == response_etag }
   end
 
   def not_modified?(request_not_modified_since:, response_last_modified:)
@@ -280,6 +282,10 @@ class Rage::Request
     return false if response_last_modified.nil?
 
     request_not_modified_since >= response_last_modified
+  end
+
+  def weak_etag(etag)
+    etag.delete_prefix("W/")
   end
 
   # @private
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -284,7 +284,7 @@ class Rage::Request
     request_not_modified_since >= response_last_modified
   end
 
-  def weak_etag(etag)
+  private def weak_etag(etag)
     etag.delete_prefix("W/")
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixed
 
+- [API] Use weak comparison for `If-None-Match` validation.
 - [Request] Treat IPv6 literals as non-domain hosts.
 - [Router] Fall back to `SERVER_NAME` when deriving exact host constraints.
 - [Cookies] Use request host fallback when resolving cookie domains.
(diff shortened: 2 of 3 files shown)


lib/rage/request.rb
```diff
@@ -282,6 +284,10 @@ def not_modified?(request_not_modified_since:, response_last_modified:)
     request_not_modified_since >= response_last_modified
   end
 
+  def weak_etag(etag)
```
rsamoilov:
```suggestion
  private def weak_etag(etag)
```

rsamoilov (commented):
Hi @RedZapdos123 

This looks great! Left just one minor suggestion.

lib/rage/request.rb
RedZapdos123: `weak_etag` was already private because it is under the private section, but I updated it to the inline private form from your suggestion.

rsamoilov: Ah, feel free to ignore my comment next time.

rsamoilov (approved):
🚀

PR #326, opened 2026-06-11
[OpenAPI]: Parse association declarations in Blueprinter

### LINKED ISSUE #291: Parse `association` declarations
Parse `association` calls, extract the `blueprint:` class reference, and recursively build a nested object schema. Support the `name:` rename option. Default to array type in the output. Fall back gracefully when the referenced blueprint cannot be resolved (emit an empty object schema, do not raise).

```
association :projects, blueprint: ProjectBlueprint
# -> "projects": { type: array, items: { ...ProjectBlueprint fields } }

association :company, blueprint: CompanyBlueprint, name: :org
# -> "org": { type: object, properties: { ... } }
```

Abishekcs:
closes #291

### What this PR does?

- Extract blueprint: class reference and recursively build nested schema
- Support name: rename option for associations
- Default to array type in the output
- Fall back to empty object schema when referenced blueprint cannot be resolved

### Example Code

- `routes.rb`

```ruby
Rage.routes.draw do
  root to: ->(env) { [200, {}, ["It works!"]] }
  get "/data_mining", to: "data_mining#show_data_mining"
end
```

- `data_mining_controller.rb`

```ruby

class DataMiningController < RageController::API

  # @response DataMining
  def show_data_mining
    data_mining_details = { 
      test_id: 'Fourth Edition',
      uuid: 'Third Edition', 
      email: 'MorganKaufmann@publishers.com', 
      first_name: 'Morgan', 
      hello_world: 'Earth', 
      projects: [
        { 
          hello_world: 'Project 1', 
          email: 'test@email.com',
          uuid: 'Third Edition',
          data_mining: [{
            uuid: 'Third Edition',
            test_id: 'Fourth Edition',
            email: 'test@publishers.com',
            first_name: 'John',
            hello_world: 'Mars'
          }]
        },
      ]
    }
    render json: DataMining.render(data_mining_details)
  end
end


```

- `data_mining_base.rb`

```ruby
  identifier :uuid
  field :hello_world
  field :email, name: :something
  association :data_mining, blueprint: DataMining
end
```

- `data_mining.rb`

```ruby
class DataMining < Blueprinter::Base
  identifier :test_id
  field "email", name: :login
  fields "first_name", :hello_world
  association :projects, blueprint: DataMiningBase, name: :classmate
end
```

- `Blueprinter Response for route /data_mining`

```
{
  "test_id": "Fourth Edition",
  "classmate": [
    {
      "uuid": "Third Edition",
      "data_mining": [
        {
          "test_id": "Fourth Edition",
          "classmate": null,
          "first_name": "John",
          "hello_world": "Mars",
          "login": "test@publishers.com"
        }
      ],
      "hello_world": "Project 1",
      "something": "test@email.com"
    }
  ],
  "first_name": "Morgan",
  "hello_world": "Earth",
  "login": "MorganKaufmann@publishers.com"
}
```

- `OpenAPI response output for # @response DataMining`

```
{
  "test_id": "string",
  "classmate": [
    {
      "uuid": "string",
      "data_mining": [
        {
          "test_id": "string",
          "classmate": [
            {
              "uuid": "string",
              "data_mining": [
                "string"
              ],
              "hello_world": "string",
              "something": "string"
            }
          ],
          "first_name": "string",
          "hello_world": "string",
          "login": "string"
        }
      ],
      "hello_world": "string",
      "something": "string"
    }
  ],
  "first_name": "string",
  "hello_world": "string",
  "login": "string"
}
```

### Screenshot

- Before 

image

- After

First

https://github.com/user-attachments/assets/a1d2e806-b857-409e-8dae-15b1157b3814

### DIFF
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -57,35 +57,25 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
   def extract_associations(reflections, view_name)
     return {} unless (view = reflections[view_name])
 
-    view.associations.each_with_object({}) do |(_, association), hash|
-      blueprint = resolve_blueprint(association.blueprint)
-      name = association.display_name.to_s
+    view.associations.each_with_object({}) do |(_, association), properties|
+      blueprint = association.blueprint
+      name = association.name.to_s
       is_collection = collection_association?(name)
 
-      if blueprint.nil?
-        item_schema = { "type" => "string" }
+      item_schema = if blueprint.is_a?(Proc)
+        { "type" => "object" }
       elsif @parsing_stack.include?(blueprint.name)
         @root.schema_registry[blueprint.name] ||= nil
-        item_schema = { "$ref" => "#/components/schemas/#{blueprint.name}" }
+        { "$ref" => "#/components/schemas/#{blueprint.name}" }
       else
-        item_schema = build_schema(blueprint, false)
+        build_schema(blueprint, false)
       end
 
-      hash[name] = is_collection ? { "type" => "array", "items" => item_schema } : item_schema
+      properties[name] = is_collection ? { "type" => "array", "items" => item_schema } : item_schema
     end
   end
 
   private
-
-  def resolve_blueprint(blueprint)
-    return blueprint unless blueprint.respond_to?(:call)
-
-    begin
-      blueprint.call(nil)
-    rescue
-      nil
-    end
-  end
 
   def collection_association?(name)
     return true unless name.respond_to?(:singularize)
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -4,6 +4,7 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
   def initialize(namespace: Object, root: Rage::OpenAPI::Nodes::Root.new, **)
     @namespace = namespace
     @root = root
+    @parsing_stack = Set.new
   end
 
   def known_definition?(str)
@@ -16,17 +17,28 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
   def parse(klass_str)
     is_collection, raw_klass_str, _ = Rage::OpenAPI.__parse_serializer_args(klass_str)
     klass = @namespace.const_get(raw_klass_str)
-    build_schema(klass, is_collection)
+    schema = build_schema(klass, is_collection)
+
+    if @root.schema_registry.key?(raw_klass_str)
+      @root.schema_registry[raw_klass_str] = is_collection ? schema["items"] : schema
+    end
+
+    schema
   end
 
   private
 
   def build_schema(klass, is_collection)
+    @parsing_stack.add(klass.name)
+
     reflections = klass.reflections
     identifier_fields = extract_fields(reflections, :identifier)
     default_fields = extract_fields(reflections, :default)
+    association_fields = extract_associations(reflections, :default)
 
-    schema = identifier_fields.merge(default_fields.sort.to_h)
+    @parsing_stack.delete(klass.name)
+
+    schema = identifier_fields.merge(default_fields.merge(association_fields).sort.to_h)
 
     result = { "type" => "object" }
     result["properties"] = schema if schema.any?
@@ -39,6 +51,27 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
 
     view.fields.each_with_object({}) do |(_, field), hash|
       hash[field.display_name.to_s] = { "type" => "string" }
+    end
+  end
+
+  def extract_associations(reflections, view_name)
+    return {} unless (view = reflections[view_name])
+
+    view.associations.each_with_object({}) do |(_, assoc), hash|
+      blueprint = assoc.blueprint
+
+      if @parsing_stack.include?(blueprint&.name)
+        @root.schema_registry[blueprint&.name] ||= nil
+        hash[assoc.display_name.to_s] = {
+          "type" => "array",
+          "items" => { "$ref" => "#/components/schemas/#{blueprint&.name}" }
+        }
+      else
+        hash[assoc.display_name.to_s] = {
+          "type" => "array",
+          "items" => build_schema(blueprint, false)
+        }
+      end
     end
   end
 end
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -59,7 +59,7 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
 
     view.associations.each_with_object({}) do |(_, association), properties|
       blueprint = association.blueprint
-      name = association.name.to_s
+      name, display_name = association.name.to_s, association.display_name.to_s
       is_collection = collection_association?(name)
 
       item_schema = if blueprint.is_a?(Proc)
@@ -71,7 +71,7 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
         build_schema(blueprint, false)
       end
 
-      properties[name] = is_collection ? { "type" => "array", "items" => item_schema } : item_schema
+      properties[display_name] = is_collection ? { "type" => "array", "items" => item_schema } : item_schema
     end
   end
 
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -57,21 +57,39 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
   def extract_associations(reflections, view_name)
     return {} unless (view = reflections[view_name])
 
-    view.associations.each_with_object({}) do |(_, assoc), hash|
-      blueprint = assoc.blueprint
+    view.associations.each_with_object({}) do |(_, association), hash|
+      blueprint = resolve_blueprint(association.blueprint)
+      name = association.display_name.to_s
+      is_collection = collection_association?(name)
 
-      if @parsing_stack.include?(blueprint&.name)
-        @root.schema_registry[blueprint&.name] ||= nil
-        hash[assoc.display_name.to_s] = {
-          "type" => "array",
-          "items" => { "$ref" => "#/components/schemas/#{blueprint&.name}" }
-        }
+      if blueprint.nil?
+        item_schema = { "type" => "string" }
+      elsif @parsing_stack.include?(blueprint.name)
+        @root.schema_registry[blueprint.name] ||= nil
+        item_schema = { "$ref" => "#/components/schemas/#{blueprint.name}" }
       else
-        hash[assoc.display_name.to_s] = {
-          "type" => "array",
-          "items" => build_schema(blueprint, false)
-        }
+        item_schema = build_schema(blueprint, false)
       end
+
+      hash[name] = is_collection ? { "type" => "array", "items" => item_schema } : item_schema
     end
+  end
+
+  private
+
+  def resolve_blueprint(blueprint)
+    return blueprint unless blueprint.respond_to?(:call)
+
+    begin
+      blueprint.call(nil)
+    rescue
+      nil
+    end
+  end
+
+  def collection_association?(name)
+    return true unless name.respond_to?(:singularize)
+
+    name.singularize != name
   end
 end
--- a/spec/support/contexts/mocked_classes.rb
+++ b/spec/support/contexts/mocked_classes.rb
@@ -29,8 +29,8 @@ RSpec.shared_context "mocked_classes" do
     if block
       if defined?(Blueprinter::Base) && parent.ancestors.include?(Blueprinter::Base)
         klass.class_eval(block.call)
-      else
-        klass.class_eval(&block)
+        Object.send(:remove_const, class_name) if Object.const_defined?(class_name, false)
+        Object.const_set(class_name, klass)
       end
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [OpenAPI] Add support for blueprint inheritance. Child blueprints now inherit fields from parent blueprints.
 - [OpenAPI] Refactor `Rage::OpenAPI::Parsers::Ext::Blueprinter` using live reflection instead of Prism AST traversal to simplify schema extraction
 - Connection pool improvements (#301).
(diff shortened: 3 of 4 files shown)


spec/support/contexts/mocked_classes.rb
```diff
@@ -29,8 +29,8 @@ class #{class_name} #{"< #{parent.name}" if parent != Object}
     if block
       if defined?(Blueprinter::Base) && parent.ancestors.include?(Blueprinter::Base)
         klass.class_eval(block.call)
-      else
-        klass.class_eval(&block)
+        Object.send(:remove_const, class_name) if Object.const_defined?(class_name, false)
```
Abishekcs, line 32: This two lines was added to deal with:

Image

Abishekcs: In test cases when using `association`

Abishekcs:
@rsamoilov PR is ready for review. Thank you!

`Test Cases Added:`

- basic association defaulting to array type with nested blueprint schema
- association name alias
- circular association fallback to $ref without infinite loop
- associations across multiple levels of inheritance
- identifier in associated blueprint with correct ordering
- circular association through inheritance
- multiple associations in one blueprint
- namespaced associations (e.g. V1::ProjectBlueprint)

`Removed test case:`

- unresolvable blueprint fallback to empty object schema

```ruby

      context "when referenced blueprint cannot be resolved" do
       let_class("UserBlueprint", parent: Blueprinter::Base) do
         <<~'RUBY'
           fields :email
           association :projects, blueprint: UnknownBlueprint
         RUBY
       end
    
       it "falls back to empty object schema" do
         is_expected.to eq({
           "type" => "object",
           "properties" => {
             "email" => { "type" => "string" },
             "projects" => {
               "type" => "array",
               "items" => { "type" => "object" }
             }
           }
         })
       end
      end

```
### question

-  Fall back gracefully when the referenced blueprint cannot be resolved (emit an empty object schema, do not raise).

Should we handle this case? Since we are no longer using parser.

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -41,4 +53,25 @@ def extract_fields(reflections, view_name)
...
+
+    view.associations.each_with_object({}) do |(_, assoc), hash|
+      blueprint = assoc.blueprint
+
+      if @parsing_stack.include?(blueprint&.name)
+        @root.schema_registry[blueprint&.name] ||= nil
+        hash[assoc.display_name.to_s] = {
+          "type" => "array",
+          "items" => { "$ref" => "#/components/schemas/#{blueprint&.name}" }
+        }
```
rsamoilov: When you use `blueprint&.name` here - how can `blueprint` be `nil`? Can this happen at all? All tests pass if I change it to `blueprint.name`.

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -41,4 +53,25 @@ def extract_fields(reflections, view_name)
...
+      else
+        hash[assoc.display_name.to_s] = {
+          "type" => "array",
```
rsamoilov: This would make all associations render as arrays. Let's check if `String#singularize` (from Active Support) is available here. If it is, we can decide if an association is an object or a collection based on `assoc.display_name.singularize != assoc.display_name`.

spec/support/contexts/mocked_classes.rb
rsamoilov: Thanks for the explanation. 

I can see the problem now - we should create a new method for Blueprinter tests. So that we have a helper to create file-based classes for Prism tests and a helper to create real classes (without polluting the namespace, ideally) for Reflection tests. 

I’ll look into it more this week.

lib/rage/openapi/parsers/ext/blueprinter.rb
Abishekcs: Okay. This helps a lot. Thank you

lib/rage/openapi/parsers/ext/blueprinter.rb
Abishekcs: Ty for pointing this out.

lib/rage/openapi/parsers/ext/blueprinter.rb
Abishekcs: >  Let's check if String#singularize (from Active Support) is available here ...

I had some doubts: since this will work only if the application has `Active Support`, if Active Support is not available, do we fall back to `type: Array`, as we won't be able to use `String#singularize`?

spec/support/contexts/mocked_classes.rb
rsamoilov: OK, I've come up with another `let_class` implementation. Feel free to leave this code as is - I'll create a PR and tag you in it once this one is merged.

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: I don't think there're many apps that don't use Active Support or Active Record.

If Active Support is not available, we can either fall back to `type: Array` or reuse [this logic](https://github.com/rage-rb/rage/blob/main/lib/rage/router/dsl.rb#L503-L526). I don't really have a strong preference.

rsamoilov: Just make sure you don't do `@active_support_loaded ||= str.respond_to?(:pluralize) || :false` like in the code I've just linked - it shouldn't have been there in the first place 😄 

Just doing `respond_to?` on demand is fine.

rsamoilov (commented):
Oh, I also found an issue - parsing currently fails with dynamic associations.

```ruby
association :project, blueprint: ->(_) { ProjectBlueprint }
```

Abishekcs:
> Oh, I also found an issue - parsing currently fails with dynamic associations.
> 
> ```ruby
> association :project, blueprint: ->(_) { ProjectBlueprint }
> ```

oh I didn't add any code to handle this my mistake.

lib/rage/openapi/parsers/ext/blueprinter.rb
Abishekcs: Okay got it. I was so confused about this.

Learning a lot. Thanks again @rsamoilov

spec/support/contexts/mocked_classes.rb
Abishekcs: ok

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -41,4 +53,43 @@ def extract_fields(reflections, view_name)
...
+
+  private
+
+  def resolve_blueprint(blueprint)
+    return blueprint unless blueprint.respond_to?(:call)
+
+    begin
+      blueprint.call(nil)
+    rescue
+      nil
```
Abishekcs: For Dealing with:



```ruby

association :project,
            name: :classmate,
            blueprint: ->(data_mining) {   
              data_mining[:subject] == 'Graph Theroy' ? DataMiningExtended : DataMiningBase
            }


```

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -41,4 +53,43 @@ def extract_fields(reflections, view_name)
...
+  def extract_associations(reflections, view_name)
+    return {} unless (view = reflections[view_name])
+
+    view.associations.each_with_object({}) do |(_, association), hash|
+      blueprint = resolve_blueprint(association.blueprint)
+      name = association.display_name.to_s
+      is_collection = collection_association?(name)
+
+      if blueprint.nil?
+        item_schema = { "type" => "string" }
```
Abishekcs: Will return just a `string` instead of `array or object` when can't resolve a Proc.

For examples like this which can't be resolve: will return `type of string`

```ruby


association :project,
            name: :classmate,
            blueprint: ->(data_mining) {   
              data_mining[:subject] == 'Graph Theroy' ? DataMiningExtended : DataMiningBase
            }



```

Abishekcs: Not sure whether this was the right choice.

Abishekcs:
New test cases added for `Dynamic association`

For both `Collection` and `Single` resource:

- [with a blueprint: Proc that ignores its argument resolves the Proc by calling it and renders the same schema as a static reference](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R702)
- [with a proc that branches on the parent object (not statically resolvable) falls back to a generic untyped item schema instead of guessing a branch](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R742)
- [with a circular association expressed as a Proc does not infinite-loop and falls back to $ref for the circular reference](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R786)
- [with a pluralized association name renders as an array](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R811)
- [with a singular association name renders as a single object, not array-wrapped](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R847)
- [with a singular association name when ActiveSupport's singularize is unavailable  falls back to array, since cardinality cannot be determined without singularize](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R880)
- [with a pluralized association name and a singular name: alias cardinality follows the alias, not the original association key ](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R920)
- [with an unresolvable dynamic blueprint on a singular association name falls back to a generic untyped schema directly (no array wrapper, since name is singular) ](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R953)
- [known limitation: uncountable noun association name correctly detects :data as a collection, since ActiveSupport singularizes it to :datum](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R990)
- [known limitation: genuinely uncountable noun association name is treated as a single object even if the real relationship is a collection](https://github.com/rage-rb/rage/pull/326/changes#diff-546f8f786704f4252d7f10a0deae3bf133f806be84d1ff79876dad99472c72a3R1025)

lib/rage/openapi/parsers/ext/blueprinter.rb
Abishekcs: In case of something like

```ruby

association :project, blueprint: ->(_) { ProjectBlueprint }

```
it will just  `blueprint.call(nil)`

Abishekcs:
@rsamoilov this PR is ready for review. Please review it whenever you are free. Thank you!

[Newly added code to deal with dynamic assocation.](https://github.com/rage-rb/rage/pull/326/changes/6d684dfe2da73e91dd6cd3f50ba1ca2365a06d23#diff-52937e657131447ddf594419ce764ea58ffde011cfedad94148fab18320d64f2)

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: > association :project, blueprint: ->(_) { ProjectBlueprint }

Thing is, I don't think it's a realistic use case. The proc expects an association object, and calling it with `nil` wil result in it failing with an exception 99% of time.

I think we need to accept that if someone uses a dynamic association, this means the blueprint is unresolvable.

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: I think `object` is a more valid choice for an association we can't resolve, which are all dynamic associations.

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -41,4 +53,43 @@ def extract_fields(reflections, view_name)
...
     end
   end
+
+  def extract_associations(reflections, view_name)
+    return {} unless (view = reflections[view_name])
+
+    view.associations.each_with_object({}) do |(_, association), hash|
+      blueprint = resolve_blueprint(association.blueprint)
+      name = association.display_name.to_s
+      is_collection = collection_association?(name)
```
rsamoilov, line 63:
```suggestion
      is_collection = collection_association?(association.name.to_s)
```

I think it's more semantically correct to use the association name as the source of truth instead of the display name.

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -41,4 +53,43 @@ def extract_fields(reflections, view_name)
       hash[field.display_name.to_s] = { "type" => "string" }
     end
   end
+
+  def extract_associations(reflections, view_name)
+    return {} unless (view = reflections[view_name])
+
+    view.associations.each_with_object({}) do |(_, association), hash|
```
rsamoilov:
```suggestion
    view.associations.each_with_object({}) do |(_, association), properties|
```

rsamoilov (commented):
Left several more comments.

Overall - amazing work!

Abishekcs:
@rsamoilov I have made the requested changes. Please have another look again, whenever you are free. Thanks again.

There are just some minor changes in test cases to reflect the new changes.

[[OpenAPI]: Use association.name instead of display_name, skip Proc blueprint resolution](https://github.com/rage-rb/rage/pull/326/changes/f4c6b841241f085a88dd2c38c7b3000ddd518c2b)

rsamoilov:
> @rsamoilov I have made the requested changes. Please have another look again, whenever you are free. Thanks again.
> 
> There are just some minor changes in test cases to reflect the new changes.
> 
> [[OpenAPI]: Use association.name instead of display_name, skip Proc blueprint resolution](https://github.com/rage-rb/rage/pull/326/changes/f4c6b841241f085a88dd2c38c7b3000ddd518c2b)

I made a [small change](https://github.com/rage-rb/rage/pull/326/commits/231888bfc6e8f9c20237dcf4863d5a961bcc07db) - what I meant was that we want the association key to be the source of truth for the collection/single object decision. We still want to use `display_name` in the schema. Let me know if you think it's wrong.

rsamoilov (approved):

Abishekcs:
I think it's okay.

> what I meant was that we want the association key to be the source of truth for the collection/single object decision.

Will keep a note to mention this while updating the docs.

PR #330, opened 2026-06-15
[API] Ignore If-Modified-Since when If-None-Match is present

### LINKED ISSUE #329: [API] Ignore If-Modified-Since when If-None-Match is present
## Description:

Verified on latest `main` at `aa15f81a245fbc9c031fcb9c6e03fcffb275c7bd`.

`Rage::Request#fresh?` combines `If-None-Match` and `If-Modified-Since` with a logical `&&`.

When `If-None-Match` matches but `If-Modified-Since` is older than the resource, Rage renders `200 OK` and returns the response body.

For requests that include `If-None-Match`, `If-Modified-Since` should be ignored.

### Minimal reproducible example:

```ruby
require "digest/sha1"
require "time"
require "rage/all"

class TestController < RageController::API
  def show
    return unless stale?(etag: "123", last_modified: Time.utc(2023, 12, 1))

    render plain: "body"
  end
end

handler = TestController.__register_action(:show)
matching_etag = %(W/"#{Digest::SHA1.hexdigest("123")}")

pp TestController.new({ "HTTP_IF_NONE_MATCH" => matching_etag }, {}).public_send(handler)

pp TestController.new({
  "HTTP_IF_NONE_MATCH" => matching_etag,
  "HTTP_IF_MODIFIED_SINCE" => Time.utc(2023, 11, 15).httpdate
}, {}).public_send(handler)
```

## Steps to reproduce:

1. Run the request with only a matching `If-None-Match` header.
2. Run the same request again with the same matching `If-None-Match` header and an older `If-Modified-Since` header.
3. Compare the two responses.

Image

### Actual behavior:

- With only matching `If-None-Match`, Rage returns `304 Not Modified`.
- With the same matching `If-None-Match` plus older `If-Modified-Since`, Rage returns `200 OK` and renders the body.
- Output on latest `main`:

```text
{:only_if_none_match=>
  [304,
   {"content-type"=>"application/json; charset=utf-8",
    "etag"=>"W/\"40bd001563085fc35165329ea1ff5c5ecbdbbeef\"",
    "last-modified"=>"Fri, 01 Dec 2023 00:00:00 GMT"},
   []],
 :if_none_match_and_older_if_modified_since=>
  [200,
   {"content-type"=>"text/plain; charset=utf-8",
    "etag"=>"W/\"40bd001563085fc35165329ea1ff5c5ecbdbbeef\"",
    "last-modified"=>"Fri, 01 Dec 2023 00:00:00 GMT"},
   ["body"]]}
```

### Expected behavior:

- When `If-None-Match` is present and matches, Rage should ignore `If-Modified-Since`.
- Both requests above should return `304 Not Modified`.

RedZapdos123:
## Description:

Issue #329 reported that Rage still evaluated `If-Modified-Since` when `If-None-Match` was present, so an older `If-Modified-Since` could force a `200 OK` even when `If-None-Match` already matched.

This PR updates `Rage::Request#fresh?` to give `If-None-Match` precedence when it is present and only fall back to `If-Modified-Since` when no `If-None-Match` header is sent.

It also adds regression tests for the matching `If-None-Match` + stale `If-Modified-Since` case and for the path where the response only sets an ETag.

Closes #329.

## Checklist:

- [x] I have run linting checks in WSL using `bundle exec rubocop --except Layout/EndOfLine lib/rage/request.rb spec/controller/api/conditional_get_spec.rb`.
- [x] I have run focused tests in WSL using `bundle exec rspec spec/controller/api/conditional_get_spec.rb`.
- [x] I have run broader regression checks in WSL using `bundle exec rspec spec/controller/api`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 
+- [API] Ignore `If-Modified-Since` when `If-None-Match` is present.
 - [Request] Treat IPv6 literals as non-domain hosts.
 - [Router] Fall back to `SERVER_NAME` when deriving exact host constraints.
 - [Cookies] Use request host fallback when resolving cookie domains.
--- a/lib/rage/request.rb
+++ b/lib/rage/request.rb
@@ -169,16 +169,18 @@ class Rage::Request
   #  request.fresh?(last_modified: Time.utc(2023, 12, 15))
   #  request.fresh?(etag: "123")
   def fresh?(etag:, last_modified:)
+    request_if_none_match = if_none_match
+    request_not_modified_since = if_not_modified_since
+
     # Always render response when no freshness information
     # is provided in the request.
-    return false unless if_none_match || if_not_modified_since
+    return false unless request_if_none_match || request_not_modified_since
 
-    etag_matches?(
-      requested_etags: if_none_match, response_etag: etag
-    ) && not_modified?(
-      request_not_modified_since: if_not_modified_since,
-      response_last_modified: last_modified
-    )
+    if request_if_none_match
+      etag_matches?(requested_etags: request_if_none_match, response_etag: etag)
+    else
+      not_modified?(request_not_modified_since: request_not_modified_since, response_last_modified: last_modified)
+    end
   end
 
   # Get the domain part of the request.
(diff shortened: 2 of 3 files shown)


rsamoilov (approved):
That's amazing!

Interestingly, Rails used to be doing [the same check](https://github.com/rails/rails/blob/v8.1.3/actionpack/lib/action_dispatch/http/cache.rb#L45-L65), but it was fixed at some point.

PR #332, opened 2026-06-18
[Request] Normalize malformed SERVER_NAME fallback hosts

### LINKED ISSUE #331: [Request] Host fallback paths do not normalize malformed SERVER_NAME values with an embedded port
## Description:

When `HTTP_HOST` is missing and `SERVER_NAME` already contains `:port`, Rage does not normalize the fallback host value before using it in request-host code paths.

This is an edge case Rack env, but it currently breaks multiple host-based paths on latest `main`.

## MRE:

```ruby
require "domain_name"
require "./lib/rage/all"

request_env = {
  "REQUEST_METHOD" => "GET",
  "rack.url_scheme" => "http",
  "PATH_INFO" => "/",
  "QUERY_STRING" => "",
  "rack.input" => StringIO.new,
  "SERVER_NAME" => "google.com:3000",
  "SERVER_PORT" => "3000"
}

request = Rage::Request.new(request_env)
puts "request.host => #{request.host.inspect}"
puts "request.url => #{request.url.inspect}"
puts "request.domain => #{request.domain.inspect}"

router = Rage::Router::Backend.new
router.on("GET", "/photos", ->(_) { "get photos" }, constraints: { host: "google.com" })
route_env = request_env.merge("PATH_INFO" => "/photos")
handler = router.lookup(route_env)
puts "router.lookup => #{handler.inspect}"

cookie_env = request_env.merge("SERVER_NAME" => "cookie.test.com:3000")
headers = {}

begin
  cookies = Rage::Cookies.new(cookie_env, headers)
  cookies[:x] = { value: 1, domain: :all }
  puts "cookies header => #{headers.inspect}"
rescue => e
  puts "cookies error => #{e.class}: #{e.message}"
end
```

Image

## Current behavior:

- `request.host` is `nil`.
- `request.url` becomes `"http://google.com:3000:3000/"`.
- `request.domain` is `nil`.
- exact host constraints do not match because `router.lookup` returns `nil`.
- cookie domain resolution can raise `TypeError: NilClass is not a String` when `domain: :all` is used.

### Actual output:

```ruby
request.host => nil
request.url => "http://google.com:3000:3000/"
request.domain => nil
router.lookup => nil
cookies error => TypeError: NilClass is not a String
```

## Expected behavior:

- Rage should normalize the fallback host the same way it normalizes `HTTP_HOST`, even if `SERVER_NAME` is malformed and already includes a port.
- `request.host` should be `"google.com"`.
- `request.url` should be `"http://google.com:3000/"`.
- `request.domain` should be `"google.com"`.
- exact host constraints for `"google.com"` should still match.
- cookie domain resolution should derive from `"cookie.test.com"` instead of raising.

RedZapdos123:
## Description:

Issue #331 reported that host fallback paths could break when `HTTP_HOST` was missing and `SERVER_NAME` already included a port, because Rage used the malformed fallback value as-is.

This PR normalizes malformed `SERVER_NAME` fallback hosts before they are used in request host handling, exact host constraints, and cookie domain resolution.

It also adds regression tests for request URL/host/domain handling, router exact host constraints, and cookie domain resolution when `SERVER_NAME` contains an embedded port, and updates `CHANGELOG.md`.

Closes #331.

## Checklist:

- [x] I have run linting checks using `bundle exec rubocop --except Layout/EndOfLine lib/rage/internal.rb lib/rage/request.rb lib/rage/cookies.rb lib/rage/router/constrainer.rb spec/rage/request_spec.rb spec/router/constraints_spec.rb spec/controller/api/cookies_spec.rb`.
- [x] I have run focused tests using `bundle exec rspec spec/rage/request_spec.rb spec/router/constraints_spec.rb spec/controller/api/cookies_spec.rb`.
- [x] I have run broader regression checks in WSL using `bundle exec rspec spec/controller/api spec/rage/request_spec.rb spec/router/constraints_spec.rb`.

### Before the fix:

image

### After the fix:

image

### DIFF
--- a/lib/rage/cookies.rb
+++ b/lib/rage/cookies.rb
@@ -201,8 +201,7 @@ class Rage::Cookies
     end
 
     if (domain = value[:domain])
-      host = Rack::Request.new(@env).host
-      host ||= Rage::Internal.extract_host(@env["SERVER_NAME"]) unless @env["HTTP_HOST"]
+      host = Rage::Request.new(@env).host
 
       processed_domain = if domain.is_a?(String)
         domain
--- a/lib/rage/cookies.rb
+++ b/lib/rage/cookies.rb
@@ -202,6 +202,7 @@ class Rage::Cookies
 
     if (domain = value[:domain])
       host = Rack::Request.new(@env).host
+      host ||= Rage::Internal.extract_host(@env["SERVER_NAME"]) unless @env["HTTP_HOST"]
 
       processed_domain = if domain.is_a?(String)
         domain
--- a/lib/rage/internal.rb
+++ b/lib/rage/internal.rb
@@ -44,6 +44,19 @@ class Rage::Internal
       }.join(", ")
     end
 
+    # Extract the host from a host:port authority while leaving bare IPv6 literals unchanged.
+    # @param authority [String, nil]
+    # @return [String, nil]
+    def extract_host(authority)
+      if authority&.start_with?("[")
+        authority.sub(/\]:\d+\z/, "]")
+      elsif authority&.count(":") == 1
+        authority.sub(/:\d+\z/, "")
+      else
+        authority
+      end
+    end
+
     # Generate a stream name based on the provided object.
     # @param streamables [#id, String, Symbol, Numeric, Array] an object that will be used to generate the stream name
     # @return [String] the generated stream name
--- a/lib/rage/internal.rb
+++ b/lib/rage/internal.rb
@@ -44,19 +44,6 @@ class Rage::Internal
       }.join(", ")
     end
 
-    # Extract the host from a host:port authority while leaving bare IPv6 literals unchanged.
-    # @param authority [String, nil]
-    # @return [String, nil]
-    def extract_host(authority)
-      if authority&.start_with?("[")
-        authority.sub(/\]:\d+\z/, "]")
-      elsif authority&.count(":") == 1
-        authority.sub(/:\d+\z/, "")
-      else
-        authority
-      end
-    end
-
     # Generate a stream name based on the provided object.
     # @param streamables [#id, String, Symbol, Numeric, Array] an object that will be used to generate the stream name
     # @return [String] the generated stream name
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - [API] Ignore `If-Modified-Since` when `If-None-Match` is present.
 - [API] Use weak comparison for `If-None-Match` validation.
+- [Request] Normalize malformed `SERVER_NAME` fallback hosts that already include a port.
 - [Request] Treat IPv6 literals as non-domain hosts.
 - [Router] Fall back to `SERVER_NAME` when deriving exact host constraints.
 - [Cookies] Use request host fallback when resolving cookie domains.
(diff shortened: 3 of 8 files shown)


rsamoilov (commented):
I'm a bit hesitant about these changes.

They seem correct overall, but in practice, Iodine rejects requests without the `Host` header before they even reach Rage. So the only way this code is likely to be executed is in some edge-case tests.

Would you suggest to still merge it?

RedZapdos123:
I think it is reasonable to merge because the code changes are narrow and only affects the existing `SERVER_NAME` fallback branch when that fallback value is malformed and already includes a port.

So....This does not change normal requests with `HTTP_HOST`. It also does not change well formed `SERVER_NAME` fallback behavior. It only makes the fallback handling internally consistent for direct Rack env entry points in Rage.

I checked a few cases against the current main and this branch:

- direct `Rage::Request.new(env)`:
  - before: `request.host => nil`, `request.url => "http://google.com:3000:3000/"`, `request.domain => nil`.
  - after: host/url/domain normalize correctly.

- Direct exact host lookup through `Rage::Router::Backend#lookup(env)`:
  - before: the host-constrained lookup returned `nil`.
  - after: it matches correctly.

- Direct `Rage::Cookies.new(env, headers)`:
  - Before: `domain: :all` raised `TypeError: NilClass is not a String`.
  - After: it resolves `domain=test.com` correctly.

- `Rage::Cable::Connection#request`:
  - Before: `cable.request.host => nil` and `cable.request.url` had the duplicated port.
  - After: both normalize correctly.

Though the real frequency of such cases maybe low, but given that the current behavior can still produce `nil` hosts, duplicated ports in `request.url`, failed exact host matches, and cookie-domain errors in that fallback branch. I think this PR is a good, low risk fix for these already existing direct Rack env code paths.

lib/rage/cookies.rb
```diff
@@ -202,6 +202,7 @@ def []=(key, value)
 
     if (domain = value[:domain])
       host = Rack::Request.new(@env).host
+      host ||= Rage::Internal.extract_host(@env["SERVER_NAME"]) unless @env["HTTP_HOST"]
```
rsamoilov: Can we instead use `Rage::Request` here?

lib/rage/internal.rb
```diff
@@ -44,6 +44,19 @@ def build_arguments(method, arguments)
       }.join(", ")
     end
 
+    # Extract the host from a host:port authority while leaving bare IPv6 literals unchanged.
+    # @param authority [String, nil]
+    # @return [String, nil]
+    def extract_host(authority)
```
rsamoilov: I think `Rage::Request` would be a better place for this method.

RedZapdos123: I moved the host normalization helper into `Rage::Request` so it stays with the request abstraction instead of `Rage::Internal`.

lib/rage/cookies.rb
RedZapdos123: Yes. Thank You for the suggestion.

I changed this to use `Rage::Request.new(@env).host` so cookies reuse the same normalized request host path instead of duplicating the fallback handling here.

rsamoilov:
Thank you for opening the PR. Unfortunately, I'm inclined to close the PR for several reasons:

* Iodine rejects requests without the Host header before they even reach Rage
* `google.com:3000` is not a valid `SERVER_NAME` according to the Rack spec
* I couldn't reproduce a real-world workflow where the fallback would work

PR #334, opened 2026-06-18
[Deferred] Ignore missing temp files during async cleanup

### LINKED ISSUE #333: [Deferred] Async temp-file cleanup can crash later reactor-based specs
## Description:

The deferred disk backend schedules temp-file cleanup with `Iodine.run_after`, and that delayed unlink can leak into later reactor-based specs.

On latest `main`, this can raise `Errno::ENOENT` from `lib/rage/deferred/backends/disk.rb` after the disk backend spec has already cleaned up its temp directory.

## Steps to reproduce:

1. Use latest `main`.
2. Run:

```bash
bundle exec rspec spec/deferred/backends/disk_spec.rb spec/logger_spec.rb
```

## Current behavior:

- `spec/deferred/backends/disk_spec.rb` passes.
- later reactor-based logger specs start Iodine.
- an async cleanup callback from the disk backend tries to unlink a temp file that is already gone.
- Iodine reports an unprotected exception from the deferred disk backend cleanup path.

### Observed output:

```text
ERROR: Iodine caught an unprotected exception - Errno::ENOENT: No such file or directory @ apply2files - /tmp/.../test_prefix...
.../lib/rage/deferred/backends/disk.rb:245:in `unlink'
.../lib/rage/deferred/backends/disk.rb:245:in `block in rotate_storage'
.../spec/support/reactor_helper.rb:16:in `within_reactor'
.../spec/logger_spec.rb:708:in `block (4 levels) in <top (required)>'
```

## Why it happens:

- `rotate_storage` schedules:

```ruby
Iodine.run_after(@fsync_frequency) do
  old_storage.close
  File.unlink(old_storage.path)
end
```

- `recover_tasks` also schedules:

```ruby
Iodine.run_after(@fsync_frequency) do
  storage.close
  File.unlink(storage.path)
end
```

- `spec/deferred/backends/disk_spec.rb` removes the temp directory in `after`.
- later reactor-based specs give those delayed callbacks a chance to run against files that were already removed.

## Expected behavior:

- deferred backend temp-file cleanup should not leak into later specs.
- cleanup should be safe if the file was already removed.
- running `spec/deferred/backends/disk_spec.rb` together with later reactor-based specs should not produce unprotected Iodine exceptions.

RedZapdos123:
## Description:

Issue #333 reported that deferred disk backend temp-file cleanup could run asynchronously after the file had already been removed, raising `Errno::ENOENT` in later reactor-based specs.

This PR updates the deferred disk backend to route async storage cleanup through a shared helper that still closes the file handle and ignores only missing-file unlink errors.

It also adds regression tests for both rotated storage cleanup and recovered storage cleanup, and updates `CHANGELOG.md`.

Fixes #333.

### Before the fix:

- Running `bundle exec rspec spec/deferred/backends/disk_spec.rb spec/logger_spec.rb` could finish the examples and then still report an unprotected Iodine exception.
- The delayed cleanup callback could try to unlink a temp file that had already been removed.

#### Example output:

```text
Finished in 2.61 seconds (files took 1.62 seconds to load)
79 examples, 0 failures

ERROR: Iodine caught an unprotected exception - Errno::ENOENT: No such file or directory @ apply2files - /tmp/.../test_prefix...
.../lib/rage/deferred/backends/disk.rb:245:in `unlink'
.../lib/rage/deferred/backends/disk.rb:245:in `block in rotate_storage'
```

### After the fix:

- The same repro command finishes cleanly.
- Async cleanup still closes the file handle, and missing temp-file unlinks no longer raise.
- Regression coverage now exercises both rotated storage cleanup and recovered storage cleanup.

#### Example output:

```text
Finished in 2.46 seconds (files took 1.43 seconds to load)
79 examples, 0 failures
```

## Checklist:

- [x] I have run linting checks using `bundle exec rubocop lib/rage/deferred/backends/disk.rb spec/deferred/backends/disk_spec.rb`.
- [x] I have run focused tests using `bundle exec rspec spec/deferred/backends/disk_spec.rb spec/logger_spec.rb`.

### DIFF
--- a/lib/rage/deferred/backends/disk.rb
+++ b/lib/rage/deferred/backends/disk.rb
@@ -241,8 +241,7 @@ class Rage::Deferred::Backends::Disk
 
     # delete the old storage ensuring the copied data has already been written to disk
     Iodine.run_after(@fsync_frequency) do
-      old_storage.close
-      File.unlink(old_storage.path)
+      cleanup_storage(old_storage)
     end
   end
 
@@ -267,8 +266,14 @@ class Rage::Deferred::Backends::Disk
     end
 
     Iodine.run_after(@fsync_frequency) do
-      storage.close
-      File.unlink(storage.path)
+      cleanup_storage(storage)
     end
+  end
+
+  def cleanup_storage(storage)
+    storage.close
+    File.unlink(storage.path)
+  rescue Errno::ENOENT
+    nil
   end
 end
--- a/lib/rage/deferred/backends/disk.rb
+++ b/lib/rage/deferred/backends/disk.rb
@@ -271,9 +271,8 @@ class Rage::Deferred::Backends::Disk
   end
 
   def cleanup_storage(storage)
+    path = storage.path
     storage.close
-    File.unlink(storage.path)
-  rescue Errno::ENOENT
-    nil
+    File.unlink(path) if File.exist?(path)
   end
 end
--- a/spec/deferred/backends/disk_spec.rb
+++ b/spec/deferred/backends/disk_spec.rb
@@ -72,16 +72,30 @@ RSpec.describe Rage::Deferred::Backends::Disk do
 
   describe "#rotate_storage" do
     let(:task) { double("Rage::Deferred::Task") }
+    let(:task_id) { backend.add(task) }
 
     before do
       allow(backend).to receive(:rotate_storage).and_call_original
-      backend.add(task)
+      task_id
       backend.instance_variable_set(:@should_rotate, true)
     end
 
     it "rotates the storage when conditions are met" do
-      backend.remove(task)
+      backend.remove(task_id)
       expect(storage_path.glob("#{prefix}*").size).to eq(2)
+    end
+
+    it "ignores missing old storage files during async cleanup" do
+      scheduled_cleanups = []
+      allow(Iodine).to receive(:run_after) { |_, &block| scheduled_cleanups << block }
+
+      old_storage = backend.instance_variable_get(:@storage)
+
+      backend.remove(task_id)
+      File.unlink(old_storage.path)
+
+      expect(scheduled_cleanups.size).to eq(1)
+      expect { scheduled_cleanups.first.call }.not_to raise_error
     end
   end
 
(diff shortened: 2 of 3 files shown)


lib/rage/deferred/backends/disk.rb
```diff
@@ -267,8 +266,14 @@ def recover_tasks(storage)
...
-      storage.close
-      File.unlink(storage.path)
+      cleanup_storage(storage)
     end
   end
+
+  def cleanup_storage(storage)
+    storage.close
+    File.unlink(storage.path)
+  rescue Errno::ENOENT
```
rsamoilov: I’m not sure we want to swallow the exception here. Can we instead attempt to delete the file only if it exists?

spec/deferred/backends/disk_spec.rb
```diff
@@ -130,6 +144,37 @@
...
+      {
+        main_storage => "main-task-id",
+        recovered_storage => "recovered-task-id"
+      }.each do |storage, task_id|
+        serialized = Marshal.dump("Rage::Deferred::Task").dump
+        entry = "add:#{task_id}:0:#{serialized}"
+        crc = Zlib.crc32(entry).to_s(16).rjust(8, "0")
+        storage.write("#{crc}:#{entry}\n")
+        storage.close
+      end
```
rsamoilov: I know there are similar examples in this file, but this looks like something the test doesn't have to know.

Is it possible to mock the storage files instead of relying on specific names and contents?

RedZapdos123: I rewrote that regression to use a mocked recovered storage object instead of creating specific storage filenames and serialized entries.

The spec now exercises the recovered cleanup path with less knowledge of the storage file format.

lib/rage/deferred/backends/disk.rb
RedZapdos123: Changed the cleanup helper to only call `File.unlink` when the path still exists, instead of rescuing `Errno::ENOENT`.

rsamoilov: I believe you haven't pushed the commit 😄

RedZapdos123: Yes, I did forget. Pushed the new commit now.

rsamoilov (approved):
Looks great!

PR #335, opened 2026-06-18
[OpenAPI]: Add support for Blueprinter view parsing - `blocks`, `include_view`, `exclude`  

### LINKED ISSUE #292: Views: `view` blocks, `include_view`, `exclude`
Parse named view blocks. Resolve `include_view` / `include_views` by compositing fields. Apply `exclude` / `excludes`. Default to the `:default` view when no view option is given. Use the `view:` option forwarded by the tag parser from #288 to select a specific view.

```ruby
view :extended do
  include_view :normal
  field :address
  exclude :last_name
end
# @response UserBlueprint(view: :extended)
# @response UserBlueprint  <- uses :default view
```

Abishekcs:
closes #292 

## What this PR does?

Extends the Rage OpenAPI Blueprinter parser to respect the `view:` option from `@response` annotations. Previously, the third return value from `__parse_serializer_args` was discarded and the parser always used the `:default` view, producing incorrect schemas for endpoints that render a non-default Blueprinter view.

## Changes

- Pass `serializer_options` from `__parse_serializer_args` into `build_schema`
- Resolve view name from `serializer_options[:view]`, falling back to `:default`
- Pass resolved view name to `extract_fields` and `extract_associations`

## Tests Added

**Single object:**
- with a named view renders base + view fields
- with no view option falls back to :default
- with a named view that includes an association
- with a named view that has a singular association
- with a view that does not exist returns empty schema
- with identifier in a named view
- with include_view in a named view
- with exclude in a named view
- with exclude on a base-level field in a named view
- with include_view and exclude combined
- with include_view, exclude, and associations combined
- with exclude on an association in a named view
- with multiple include_view chained
- with include_view chained and exclude applied to an included field

**Collection:**
- Same 14 cases as above for `Array<Blueprint(view: :name)>` resource shape

## Screnshoot

image

### DIFF
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Rage::OpenAPI::Parsers::Ext::Blueprinter
+  class InvalidViewError < StandardError; end
+
   def initialize(namespace: Object, root: Rage::OpenAPI::Nodes::Root.new, **)
     @namespace = namespace
     @root = root
@@ -24,6 +26,8 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
     end
 
     schema
+  rescue InvalidViewError => e
+    Rage::OpenAPI.__log_warn e.message
   end
 
   private
@@ -33,10 +37,12 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
 
     view_name = serializer_options&.key?(:view) ? serializer_options[:view] : :default
     reflections = klass.reflections
+    view = reflections[view_name]
+    raise InvalidViewError, "invalid view #{view_name}" unless view
 
-    identifier_fields = extract_fields(reflections, :identifier)
-    default_fields = extract_fields(reflections, view_name)
-    association_fields = extract_associations(reflections, view_name)
+    identifier_fields = extract_fields(reflections[:identifier])
+    default_fields = extract_fields(view)
+    association_fields = extract_associations(view)
 
     @parsing_stack.delete(klass.name)
 
@@ -48,17 +54,13 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
     result
   end
 
-  def extract_fields(reflections, view_name)
-    return {} unless (view = reflections[view_name])
-
-    view.fields.each_with_object({}) do |(_, field), hash|
-      hash[field.display_name.to_s] = { "type" => "string" }
+  def extract_fields(view)
+    view.fields.each_with_object({}) do |(_, field), properties|
+      properties[field.display_name.to_s] = { "type" => "string" }
     end
   end
 
-  def extract_associations(reflections, view_name)
-    return {} unless (view = reflections[view_name])
-
+  def extract_associations(view)
     view.associations.each_with_object({}) do |(_, association), properties|
       blueprint = association.blueprint
       name, display_name = association.name.to_s, association.display_name.to_s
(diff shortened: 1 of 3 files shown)


Abishekcs:
@rsamoilov PR is ready for review. Thanks

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -15,9 +15,9 @@ def known_definition?(str)
...
 
   def parse(klass_str)
-    is_collection, raw_klass_str, _ = Rage::OpenAPI.__parse_serializer_args(klass_str)
+    is_collection, raw_klass_str, serializer_options = Rage::OpenAPI.__parse_serializer_args(klass_str)
     klass = @namespace.const_get(raw_klass_str)
-    schema = build_schema(klass, is_collection)
+    schema = build_schema(klass, is_collection, serializer_options)
 
     if @root.schema_registry.key?(raw_klass_str)
       @root.schema_registry[raw_klass_str] = is_collection ? schema["items"] : schema
```
rsamoilov, line 25:
```suggestion
if @root.schema_registry.key?([raw_klass_str, serializer_options])
@root.schema_registry[[raw_klass_str, serializer_options]] = is_collection ? schema["items"] : schema
```

Now that we parse serializer options too, should we update the schema registry?

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: Now that `view_name` is dynamic, I feel it's better to change this condition.

The `return` here silently ignores incorrect view names, but we don't want to silently ignore them - we want to fail fast and log a warning.

`Rage::OpenAPI::Parser` already logs a warning if a request parser returns `nil` - we just need to support it here.

Here's what I'd propose:

* define an internal error in this class, e.g. `InvalidViewError`
* if the view is not found - `raise InvalidViewError`
* catch `InvalidViewError` in `parse` and return `nil`

This will make `Rage::OpenAPI::Parser` to log `unrecognized @request tag detected`. Additionally, we could also log the reason for the error (e.g. `invalid view #{view}`) when raising `InvalidViewError`.

rsamoilov: Same here - it's better to explicitly fail to parse the blueprinter instead of ignoring an incorrect view.

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: I feel it makes sense, but feel free to leave it as is if you can't reproduce the issue in tests.

Abishekcs: I can try this in a different PR ?

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -15,26 +17,32 @@ def known_definition?(str)
...
-  def build_schema(klass, is_collection)
+  def build_schema(klass, is_collection, serializer_options = nil)
     @parsing_stack.add(klass.name)
 
+    view_name = serializer_options&.key?(:view) ? serializer_options[:view] : :default
     reflections = klass.reflections
-    identifier_fields = extract_fields(reflections, :identifier)
-    default_fields = extract_fields(reflections, :default)
-    association_fields = extract_associations(reflections, :default)
+    view = reflections[view_name]
```
Abishekcs, line 40: Moved this  here instead of passing both `reflections` and `view_name` to `extract_fields` and `extract_association` and raising the error in `build_schema` itself if view not available.

Abishekcs:
@rsamoilov i have made the required changes. Please have a look when you are free. Thanks again!

lib/rage/openapi/parsers/ext/blueprinter.rb
rsamoilov: Sure!

rsamoilov (approved):
👍

PR #336, opened 2026-06-23
Update Blueprinter specs to use `let_blueprinter_class`

rsamoilov:
This pull request introduces `let_blueprinter_class` to create Blueprinter mock classes. Since we don't use Prism parsing with Blueprinter, the test setup has to be quite different from the Alba setup - we need to evaluate the classes to ensure Blueprinter's reflection API works correctly.

`let_blueprinter_class` builds class instances for tests, and supports namespaces and circular references.


rsamoilov:
@Abishekcs This is the change for the test setup I mentioned earlier. Could you please give it a look?

Abishekcs:
Went through it. Looking good. Thanks for the help.

PR #340, opened 2026-06-30
[OpenAPI]: Fix blueprinter registry key normalization and chose correct view when explicitly mentioned for association

Abishekcs:
### What this PR does ?

If the same Blueprinter class shows up under different views across a spec, say `UserBlueprint(view: :normal)` on one endpoint and `UserBlueprint(view: :extended)` on another, the OpenAPI parser used to track both under the same key: just the class name. That caused two bugs:

1. **One view's schema overwrote the other's.** Whichever view got parsed last "won" in the schema registry, so a `$ref` meant to point at one view's schema could end up resolving to a different view's schema entirely.
2. **False circular-reference detection.** The parser also used bare class names to detect cycles, so it couldn't tell *which* view of a class was already being built, only that the class name was somewhere on the stack.

It's common for one class to have *multiple* associations to the same target class, each pinned to a different view. For example: `# @response DataMining(view: :sample)`

```ruby
class DataMining < Blueprinter::Base
  fields "first_name", :hello_world
  association :project, view: :sample, blueprint: DataMiningBase, name: :classmate
  view :extended do
    field :extended_author
  end
end

  field :first_name
  association :data_mining, blueprint: DataMining, view: :extended
  view :sample do
    field :hello_world
    association :data_mining, blueprint: DataMining
  end
end
```

Parsing `DataMining` (default view): it associates to `DataMiningBase(view: :sample)`, whose own association loops back to `DataMining`'s **default** view a genuine cycle. Meanwhile, `DataMiningBase`'s *default* view separately associates to `DataMining(view: :extended)` a completely different edge that should never be touched during this parse. With the new implementation the parser will ignore `DataMiningBase(view: :extended)` as `DataMining` points to `sample` view.

### Screenshot

- Before

image

- Afrer

image

### DIFF
--- a/lib/rage/openapi/parsers/ext/blueprinter.rb
+++ b/lib/rage/openapi/parsers/ext/blueprinter.rb
@@ -20,9 +20,10 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
     is_collection, raw_klass_str, serializer_options = Rage::OpenAPI.__parse_serializer_args(klass_str)
     klass = @namespace.const_get(raw_klass_str)
     schema = build_schema(klass, is_collection, serializer_options)
+    registry_key = "#{raw_klass_str}_#{serializer_options[:view] || :default}"
 
-    if @root.schema_registry.key?(raw_klass_str)
-      @root.schema_registry[raw_klass_str] = is_collection ? schema["items"] : schema
+    if @root.schema_registry.key?(registry_key)
+      @root.schema_registry[registry_key] = is_collection ? schema["items"] : schema
     end
 
     schema
@@ -32,19 +33,19 @@ class Rage::OpenAPI::Parsers::Ext::Blueprinter
 
   private
 
-  def build_schema(klass, is_collection, serializer_options = nil)
-    @parsing_stack.add(klass.name)
-
+  def build_schema(klass, is_collection, serializer_options)
     view_name = serializer_options&.key?(:view) ? serializer_options[:view].to_sym : :default
     reflections = klass.reflections
     view = reflections[view_name]
     raise InvalidViewError, "invalid view #{view_name}" unless view
 
+    @parsing_stack.add([klass.name, view_name])
+
     identifier_fields = extract_fields(reflections[:identifier])
     default_fields = extract_fields(view)
     association_fields = extract_associations(view)
 
-    @parsing_stack.delete(klass.name)
+    @parsing_stack.delete([klass.name, view_name])
 
     schema = identifier_fields.merge(default_fields.merge(association_fields).sort.to_h)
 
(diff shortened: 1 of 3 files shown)


Abishekcs:
This PR adds support for associations that explicitly specify a `view`: option for their `blueprint`. It also adds `serializer_options` to the schema registry without this, if two associations reference the same `blueprint` but only one of them specifies a `view`, they would collide in the registry under the same key, even though they should resolve to different schemas.

Abishekcs:
@rsamoilov should we support explicitly specifying a `view:` for a blueprint association? Right now it always builds against the target's default view, even if view: is passed.

```ruby
   association :all_projects, blueprint: ProjectBlueprint, view: :normal
```

Abishekcs:
In case we decide to support it. Then we have to handle the case where in `circular association` both the association have `view` explicitly mentioned.

Example: `# @response UserBlueprint(view: :extended)`
```ruby
class ProjectBlueprint < Blueprinter::Base
  fields :name
  view :normal do
    association :users, blueprint: UserBlueprint, view: :extra_information
  end
end

class UserBlueprint < Blueprinter::Base
  fields :email
  view :extended do
    association :all_projects, blueprint: ProjectBlueprint, view: :normal
  end
  
  view :extra_information do
    field :address
  end
end
```

rsamoilov:
> @rsamoilov should we support explicitly specifying a `view:` for a blueprint association? Right now it always builds against the target's default view, even if view: is passed.
> 
> ```ruby
>    association :all_projects, blueprint: ProjectBlueprint, view: :normal
> ```

I'd say yes.

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -20,9 +20,10 @@ def parse(klass_str)
     is_collection, raw_klass_str, serializer_options = Rage::OpenAPI.__parse_serializer_args(klass_str)
     klass = @namespace.const_get(raw_klass_str)
     schema = build_schema(klass, is_collection, serializer_options)
+    registry_key = "#{raw_klass_str}_#{serializer_options[:view] || :default}"
```
Abishekcs, line 23: I have the made the key a combination of  `"#{ClassName}_#{and its view}"`.

lib/rage/openapi/parsers/ext/blueprinter.rb
```diff
@@ -32,19 +33,19 @@ def parse(klass_str)
...
     raise InvalidViewError, "invalid view #{view_name}" unless view
 
+    @parsing_stack.add([klass.name, view_name])
+
     identifier_fields = extract_fields(reflections[:identifier])
     default_fields = extract_fields(view)
     association_fields = extract_associations(view)
 
-    @parsing_stack.delete(klass.name)
+    @parsing_stack.delete([klass.name, view_name])
```
Abishekcs, line 48: Same in `@parsing_stack` this helps in choosing the `view` when view is mentioned explicitly for an association.

Abishekcs:
@rsamoilov PR is ready for review. Thanks!

rsamoilov (approved):
👍

PR #359, opened 2026-07-22
[CI][Security] Declare least-privilege GITHUB_TOKEN permissions

### LINKED ISSUE #358: [CI][Security] Declare least-privilege GITHUB_TOKEN permissions
## Description:

Verified on latest `main` at `65ed10bf37e6404df9dd4b31230bb4addcc93629`.

Four GitHub Actions workflows do not declare explicit `GITHUB_TOKEN` permissions:

- `.github/workflows/main.yml`.
- `.github/workflows/ext.yml`.
- `.github/workflows/docs.yml`.
- `.github/workflows/changelog.yml`.

Without a `permissions:` declaration, these workflows inherit the default permissions configured for the repository or organization. This makes their effective permissions depend on external settings and allows them to receive broader access if those defaults change.

The release workflow already declares its required permissions explicitly.

## Steps to reproduce:

1. Check out latest `main`.
2. Run the following command from the repository root.

```bash
ruby -e 'Dir[".github/workflows/*.yml"].sort.each { |file| declared = File.readlines(file).any? { |line| line.match?(/^\s*permissions:/) }; puts "#{file}: #{declared ? "declared" : "missing"}" }'
```

3. Review the reported workflow files.

## Current behavior:

```text
.github/workflows/changelog.yml: missing
.github/workflows/docs.yml: missing
.github/workflows/ext.yml: missing
.github/workflows/main.yml: missing
.github/workflows/release.yml: declared
```

- Test, extension, documentation, and changelog workflows inherit repository or organization defaults.
- Their required token access is not documented in the workflow files.
- A future change to the default token setting could grant these workflows more access than they need.

## Expected behavior:

- Each workflow should explicitly grant only the `GITHUB_TOKEN` permissions it requires.
- `main.yml`, `ext.yml`, and `docs.yml` should use `contents: read`.
- `changelog.yml` should use `pull-requests: read` because it reads the pull request file list.
- All unspecified permissions should remain disabled.

## Security impact:

This is a defense-in-depth change. It does not assume that the workflows currently receive write access, because the repository or organization default may already be read-only.

Explicit declarations keep the workflows least-privileged if external defaults change. They also limit the access available through `github.token`, which actions can use even when the token is not passed as an explicit input.

GitHub recommends granting the minimum required permissions to `GITHUB_TOKEN`:

- https://docs.github.com/en/actions/reference/security/secure-use#principle-of-least-privilege.
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions.

GitHub CodeQL also provides a high-precision query for workflows that do not declare permissions:

- https://codeql.github.com/codeql-query-help/actions/actions-missing-workflow-permissions/.

RedZapdos123:
## Description:

Issue #358 reported that the test, extension, documentation, and changelog workflows inherited their `GITHUB_TOKEN` permissions from repository or organization defaults.

This PR gives the test, extension, and documentation workflows read only contents access. It gives the changelog workflow read only pull request access.

This keeps all other token permissions disabled and leaves the existing release workflow permissions unchanged.

Closes #358.

## Checklist:

- [x] I have validated all workflow files using `go run github.com/rhysd/actionlint/cmd/actionlint@latest`.
- [x] I have validated the workflow YAML files using Ruby's YAML parser.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled]
 
+permissions:
+  pull-requests: read
+
 jobs:
   check:
     runs-on: ubuntu-latest
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,6 +5,9 @@ on:
     tags:
       - '*'
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/ext.yml
+++ b/.github/workflows/ext.yml
@@ -3,6 +3,9 @@ name: Ext Tests
 on:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   appraise:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,9 @@ on:
 
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest


RedZapdos123:
Since this is a CI only security hardening change. It does not affect Rage runtime behavior, public API, or any user facing functionality, so no changelog entry added.

rsamoilov (approved):
Looks good, thank you! And sorry for the delays with reviews.

PR #360, opened 2026-07-29
[WIP] Telemetry Scaffold GC pressure health check

Abishekcs:
### Work In Progress

### DIFF
--- a/lib/rage/configuration.rb
+++ b/lib/rage/configuration.rb
@@ -1049,6 +1049,14 @@ class Rage::Configuration
   # @see Rage::Configuration::MiddlewareRegistry
   # @see Rage::Telemetry
   class Telemetry < MiddlewareRegistry
+    attr_accessor :gc_pressure_monitoring_enabled
+
+    # @private
+    def initialize
+      @gc_pressure_monitoring_enabled = false
+      super
+    end
+
     # @private
     # @return [Hash{String => Array<Rage::Telemetry::HandlerRef>}] a map of span IDs to handler references
     def handlers_map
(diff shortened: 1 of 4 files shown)


rsamoilov:
Hey @Abishekcs 

I'm super excited to see the first steps here!

Do you think it makes sense to define what the end result would look like first?

The reason I'm asking is because I thought it would make sense to decouple Rage from metrics. My vision was that:

* `Rage::Telemetry` would expose generic helpers to collect metrics (e.g. `Rage::Telemetry.every(ms) {}`) and some framework-specific metrics (e.g. `Rage::Telemetry.raw[:backlog_size]`)
* the metrics like GC pressure or request count would instead be collected by a specific instrumentation, e.g. [OpenTelemetry](https://github.com/rage-rb/opentelemetry-instrumentation).

WDYT?

Abishekcs:
> Do you think it makes sense to define what the end result would look like first?

Sure, I will open a github issue this weekend  for this telemetry project laying out everything.





> The reason I'm asking is because I thought it would make sense to decouple Rage from metrics. My vision was that:
> 
> * `Rage::Telemetry` would expose generic helpers to collect metrics (e.g. `Rage::Telemetry.every(ms) {}`) and some framework-specific metrics (e.g. `Rage::Telemetry.raw[:backlog_size]`)
> * the metrics like GC pressure or request count would instead be collected by a specific instrumentation, e.g. [OpenTelemetry](https://github.com/rage-rb/opentelemetry-instrumentation).
> 
> WDYT?

Sounds good, I'll rework the scaffolding into a generic Rage::Telemetry and move the GC-specific logic over to opentelemetry-instrumentation, matching the existing handler pattern.

I've been going through GC.stat, planning to track deltas of :time (% of interval), minor_gc_count/major_gc_count, and heap occupancy still figuring out which of these actually signal real GC pressure. will try to read through some blog post about ruby gc this weekend.

Abishekcs:
Closing this will open a github issue soon as mentioned above.

PR #378, opened 2026-08-02
[Router] Remove unused custom constraint scaffold

RedZapdos123:
## Description:

This PR removes unused router custom constraint scaffold.

The router only supports host constraints as of the latest main. `Rage::Router::Backend` always builds `Rage::Router::Constrainer` without any custom strategies in practice, and the constrainer never uses the constructor argument.

This PR removes the unused constructor path and dead helper without changing router behavior.

## Checklist:

- [x] I have run linting checks using `bundle exec rubocop --format simple lib/rage/router/constrainer.rb lib/rage/router/backend.rb`.
- [x] I have run focused tests using `bundle exec rspec spec/router`.

### DIFF
--- a/lib/rage/router/backend.rb
+++ b/lib/rage/router/backend.rb
@@ -11,7 +11,7 @@ class Rage::Router::Backend
   def initialize
     @routes = []
     @trees = {}
-    @constrainer = Rage::Router::Constrainer.new({})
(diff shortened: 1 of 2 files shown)


RedZapdos123:
This is an internal router cleanup with no runtime behavior or public API change, so no need to add a changelog entry.

PR #381, opened 2026-08-04
[Telemetry] Add `Rage::Telemetry.every` for periodic block scheduling

Abishekcs:
### What this PR does?

- `Rage::Telemetry.every` now schedules each tick via `Fiber.schedule` instead of calling the block inline in the root fiber

### Why this design?

- Iodine timers fire in the root fiber, which blocks the reactor thread
- Inside a scheduler fiber, the block's I/O hooks into the fiber scheduler and yields to the reactor, keeping the server responsive

### DIFF
--- a/lib/rage/telemetry/telemetry.rb
+++ b/lib/rage/telemetry/telemetry.rb
@@ -36,27 +36,23 @@ module Rage::Telemetry
     __registry.keys
   end
 
-  # The block is executed on the reactor thread and is yielded the scheduling lag in milliseconds -
-  # the difference between the expected and actual execution times. Keep the block fast and non-blocking.
+  # Registers a block to be executed repeatedly at a fixed interval while the server is running.
+  # The block runs on the reactor thread, alongside every fiber. Keep it fast and non-blocking.
+  # If the server is not running yet, execution is deferred and starts once the server starts.
   #
   # @param interval_ms [Integer] the execution interval in milliseconds
-  # @yield [Integer] the scheduling lag in milliseconds
-  # @example Periodically record the event loop lag
-  #   Rage::Telemetry.every(100) do |lag|
-  #     MyMetrics.record_loop_lag(lag)
+  # @example Periodically sample GC stats
+  #   Rage::Telemetry.every(1000) { MyMetrics.record_gc_stats(GC.stat) }
+  #
+  # @example Measure event loop lag
+  #   last_run_at = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+  #   Rage::Telemetry.every(100) do
+  #     now = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+  #     MyMetrics.record_loop_lag([now - last_run_at - 100, 0].max)
+  #     last_run_at = now
   #   end
   def self.every(interval_ms, &block)
-    schedule_block = -> do
-      last_run_time = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
-
-      Iodine.run_every(interval_ms) do
-        current_run_time = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
-        lag_ms = [current_run_time - last_run_time - interval_ms, 0].max
-
-        block.call(lag_ms)
-        last_run_time = current_run_time
-      end
-    end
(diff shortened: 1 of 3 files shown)


Abishekcs:
@rsamoilov PR is ready for review.

Just a short summary:

I added a block to `Rage::Telemetry.every(ms, &block)` so that `opentelemetry-instrumentation` can use this to call the respective block of code they want to run at their respective interval. This block will also serve to calculate the delay in event loop lag.


Also, I wrote a test code which passes locally but kept failing in the CI `I have removed it for now`

```

  context "within the reactor" do
      before do
        Fiber.set_scheduler(Rage::FiberScheduler.new)
      end
      after do
        Fiber.set_scheduler(nil)
      end

      it "periodically executes the block" do
        counter = 0
        within_reactor do
          described_class.every(50) { counter += 1 }
          deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + 2
          sleep 0.05 while counter < 3 && Process.clock_gettime(Process::CLOCK_MONOTONIC) < deadline
          -> { expect(counter).to be >= 3 }
        end
      end
...


```

rsamoilov:
Sorry for the confusion here.

We need to distinguish between `Rage::Telemetry.every` and event loop lag. `every` should be a generic method akin to `setInterval` in JS. It can be use for anything recurring - sampling values, batching and flushing metrics, heartbeats, aggregations, event loop lag. Basically, it should only be wrapping `Iodine.run_every`, exposing its functionality as a public API.

Abishekcs:
@rsamoilov made the required changes. please have another look whenever you are free. thank you!

lib/rage/telemetry/telemetry.rb
```diff
@@ -36,6 +36,31 @@ def self.available_spans
...
+  #     last_run_at = now
+  #   end
+  def self.every(interval_ms, &block)
+    schedule_block = -> { Iodine.run_every(interval_ms, &block) }
+
+    if Iodine.running?
+      schedule_block.call
+    else
+      Iodine.on_state(:on_start, &schedule_block)
+    end
```
rsamoilov: This can be simplified to `Iodine.run_every(interval_ms, &block)` - `run_every` remembers the blocks if the server is not running and executes them once the server is ready.

lib/rage/telemetry/telemetry.rb
```diff
@@ -36,6 +36,31 @@ def self.available_spans
...
+  #   Rage::Telemetry.every(1000) { MyMetrics.record_gc_stats(GC.stat) }
+  #
+  # @example Measure event loop lag
+  #   last_run_at = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+  #   Rage::Telemetry.every(100) do
+  #     now = Process.clock_gettime(Process::CLOCK_MONOTONIC, :millisecond)
+  #     MyMetrics.record_loop_lag([now - last_run_at - 100, 0].max)
+  #     last_run_at = now
+  #   end
+  def self.every(interval_ms, &block)
```
rsamoilov, line 46: Let's also wrap blocks in fibers. 

`run_every` blocks run in the root fiber, which is blocking. Some of the timers will likely be doing I/O (e.g. metric flushing), and having them block the server will be surprising.

rsamoilov (dismissed):
Great start!

There are several comments, but I've approved the PR. Feel free to either continue here or merge it and tackle those in another one.

lib/rage/telemetry/telemetry.rb
Abishekcs: > run_every remembers the blocks if the server is not running and executes them once the server is ready.

Interesting is it because `block` of `run_every` is assigned to a  `queue`  which is part of the event loop. So, when the server starts the `event_loop` takes care of running it?

Abishekcs: I hope the question makes sense ? I am trying to understand the `Iodine` codebase.

rsamoilov: > Interesting is it because `block` of `run_every` is assigned to a `queue` which is part of the event loop. So, when the server starts the `event_loop` takes care of running it?

Exactly.

Abishekcs:
@rsamoilov made the required changes in this PR itself. Please have another look whenever you are free. Thank you!

rsamoilov (approved):
Beautiful!

PR #386, opened 2026-08-17
[Logger] Add structured log context redaction

### LINKED ISSUE #370: [Logger] Redact sensitive information from logs
### Description

Rage's [structured logging system](https://rage-rb.dev/docs/logging) makes it easy to add context to log entries using key-value pairs. However, there is currently no built-in mechanism to prevent sensitive information — such as passwords, tokens, credit card numbers, or personal data — from being written to logs.

For example, a developer might accidentally log sensitive data through `with_context` or `append_info_to_payload`:

```ruby
Rage.logger.info "user signed in", email: user.email, token: session_token
```

Or sensitive parameters might leak through request logs if a controller doesn't properly filter them.

A redaction setting would allow users to define matching rules for keys whose values should be automatically replaced with `[REDACTED]` (or similar) before the log entry is written.

Example:

```ruby
Rage.configure do
  config.logger.redact_keys = [:password, :token, :secret, :authorization, :credit_card]
end
```

Since Rage logs are structured as key-value pairs, the implementation boils down to iterating the log context and replacing values for matching keys. The more interesting design question is how the matching rules should work.

### Design considerations

- **Exact vs. pattern matching.** Should keys match exactly (`:password` matches only `password`) or should they support substring/pattern matching (`:password` also matches `password_confirmation`, `old_password`, etc.)? Substring matching is more forgiving but could produce false positives. Consider whether the default should be substring matching with an option for exact matching, or vice versa.
- **Nested keys.** If log context contains nested hashes (e.g. `{ user: { email: "...", password: "..." } }`), should redaction apply recursively? This adds complexity but is important for safety.
- **Custom matchers.** Beyond key names, some use cases require value-based redaction (e.g. redacting strings that look like credit card numbers regardless of the key name). Consider whether custom matcher procs should be supported in the initial implementation or deferred to a follow-up.
- **Request log integration.** Redaction should apply consistently across all log entry points — `Rage.logger.info` with context, `append_info_to_payload`, global `log_context`.
- **Performance.** Redaction runs on every log entry, so the matching logic should be fast.

### Tips

- Review the [Logging docs](https://rage-rb.dev/docs/logging) to understand how structured logging, `with_context`, `append_info_to_payload`, and external loggers work.
- Look at how `Rage::Logger#rebuild!` and `Rage::LogProcessor#rebuild!` use dynamic code generation — this may be a good place to integrate redaction logic efficiently.
- Check the [architecture doc](https://github.com/rage-rb/rage/blob/main/ARCHITECTURE.md) to see how Rage's core components interact and to understand the design principles.
- Read the [contributing guide](https://github.com/rage-rb/rage/blob/main/CONTRIBUTING.md) for coding conventions and design principles used across the codebase.
- **Before starting the implementation, please share your proposed design approach.** The matching rules for keys need careful designing, so discussing the approach early will drastically increase the chances of acceptance and help avoid rework.
- **Feel free to ask any questions or request help in the comments below!**

RedZapdos123:
## Description:

Issue #370 asked for a built-in way to redact sensitive values from Rage's structured log context.

This PR adds config.log_redact_keys=, which applies Rails-style case-insensitive partial matching to structured context keys and replaces matching values with [REDACTED] before the log entry is written.

The implementation stays inside Rage's own logger code and applies consistently to with_context, inline logger context, request log context added through append_info_to_payload, global config.log_context, and dynamic external logger context payloads.

As discussed in the issue, this PR only redacts structured context. It does not change request_info.

It also adds regression specs for logger redaction, configuration wiring, nested structured context, and the unchanged request_info behavior for dynamic external loggers.

Closes #370.

## Checklist:

- [x] I have added relevant regression test cases for this change.
- [x] I have run linting checks using bundle exec rubocop lib/rage/logger/logger.rb lib/rage/configuration.rb spec/logger_spec.rb spec/configuration_spec.rb.
- [x] I have run focused specs using bundle exec rspec spec/logger_spec.rb spec/configuration_spec.rb.
- [x] I have run syntax checks using bundle exec ruby -c lib/rage/logger/logger.rb && bundle exec ruby -c lib/rage/configuration.rb.

### Before the fix:

Sensitive keys in structured log context were written as-is.

Rage did not have a built-in redaction mechanism for with_context, inline logger context, or request log context merged through append_info_to_payload.

Example:

    [my_test_tag] timestamp=... level=info password_confirmation=secret AUTH_TOKEN=abc user_id=12345 message=passed

### After the fix:

config.log_redact_keys= redacts matching structured context keys before log emission.

Matching is case-insensitive and partial, so keys like password_confirmation and AUTH_TOKEN are redacted.

Nested hashes and arrays are redacted recursively.

Dynamic external loggers receive redacted context and unchanged request_info.

Example:

    [my_test_tag] timestamp=... level=info password_confirmation=[REDACTED] AUTH_TOKEN=[REDACTED] user_id=12345 message=passed

### DIFF
--- a/lib/rage/configuration.rb
+++ b/lib/rage/configuration.rb
@@ -256,6 +256,25 @@ class Rage::Configuration
   def log_tags
     @log_tags ||= LogTags.new
   end
+
+  # Allows configuring case-insensitive partial matches for filtering structured log context keys.
+  # Matching keys will have their values replaced with `"[FILTERED]"` before the log entry is written.
+  # @return [Rage::Configuration::FilterParameters]
+  #
+  # @example Filter common secrets from structured logs
+  #   Rage.configure do
+  #     config.filter_parameters = [:password, :token, :secret]
+  #   end
+  def filter_parameters
+    @filter_parameters ||= FilterParameters.new
+  end
+
+  # Replace the current list of filtered log parameter keys.
+  # @param parameters [String, Symbol, Array<String, Symbol>, nil] one or more parameter filters
+  def filter_parameters=(parameters)
+    @filter_parameters = FilterParameters.new
+    @filter_parameters << parameters if parameters
+  end
   # @!endgroup
 
   # @!group Telemetry Configuration
@@ -407,6 +426,18 @@ class Rage::Configuration
         obj.each { |item| validate_input!(item) }
       elsif !obj.respond_to?(:to_str) && !obj.respond_to?(:call)
         raise ArgumentError, "custom log tag has to be a string, an array of strings, or respond to `#call`"
+      end
+    end
+  end
+
+  class FilterParameters < LogContext
+    private
+
+    def validate_input!(obj)
+      if obj.is_a?(Array)
+        obj.each { |item| validate_input!(item) }
+      elsif !obj.is_a?(String) && !obj.is_a?(Symbol)
+        raise ArgumentError, "filter parameters have to be strings, symbols, or arrays of strings and symbols"
       end
     end
   end
@@ -1276,6 +1307,10 @@ class Rage::Configuration
     if @log_tags
       Rage.__log_processor.add_custom_tags(@log_tags.objects)
       @logger.dynamic_tags = Rage.__log_processor.dynamic_tags
+    end
+
+    if @filter_parameters
+      @logger.filter_parameters = @filter_parameters.objects
     end
 
     if before_boot && @blocking_operation_pool&.enabled
--- a/lib/rage/configuration.rb
+++ b/lib/rage/configuration.rb
@@ -73,6 +73,8 @@ class Rage::Configuration
     else
       raise ArgumentError, "Invalid logger: must be an instance of `Rage::Logger`, respond to `#call`, or implement all standard Ruby Logger methods (`#debug`, `#info`, `#warn`, `#error`, `#fatal`, `#unknown`)"
     end
+
+    @logger.log_redact_keys = @log_redact_keys if @logger && defined?(@log_redact_keys)
   end
 
   # Returns the log formatter used by Rage.
@@ -257,25 +259,40 @@ class Rage::Configuration
     @log_tags ||= LogTags.new
   end
 
-  # Allows configuring case-insensitive partial matches for filtering structured log context keys.
-  # Matching keys will have their values replaced with `"[FILTERED]"` before the log entry is written.
-  # @return [Rage::Configuration::FilterParameters]
+  # Allows configuring case-insensitive partial matches for redacting structured log context keys.
+  # Matching keys will have their values replaced with `"[REDACTED]"` before the log entry is written.
   #
-  # @example Filter common secrets from structured logs
+  # @param keys [String, Symbol, Array<String, Symbol>, nil] one or more keys to redact
+  # @example Redact common secrets from structured logs
   #   Rage.configure do
-  #     config.filter_parameters = [:password, :token, :secret]
+  #     config.log_redact_keys = [:password, :token, :secret]
   #   end
-  def filter_parameters
-    @filter_parameters ||= FilterParameters.new
-  end
-
-  # Replace the current list of filtered log parameter keys.
-  # @param parameters [String, Symbol, Array<String, Symbol>, nil] one or more parameter filters
-  def filter_parameters=(parameters)
-    @filter_parameters = FilterParameters.new
-    @filter_parameters << parameters if parameters
+  def log_redact_keys=(keys)
+    @log_redact_keys = normalize_log_redact_keys(keys)
+    @logger.log_redact_keys = @log_redact_keys if @logger
   end
   # @!endgroup
+
+  private def normalize_log_redact_keys(keys)
+    return nil if keys.nil?
+
+    validate_log_redact_keys!(keys)
+
+    normalized_keys = Array(keys).flatten.filter_map do |key|
+      key = key.to_s
+      key unless key.empty?
+    end.uniq
+
+    normalized_keys.any? ? normalized_keys.freeze : nil
+  end
+
+  private def validate_log_redact_keys!(keys)
+    if keys.is_a?(Array)
+      keys.each { |key| validate_log_redact_keys!(key) }
+    elsif !keys.is_a?(String) && !keys.is_a?(Symbol)
+      raise ArgumentError, "log redact keys have to be strings, symbols, or arrays of strings and symbols"
+    end
+  end
 
   # @!group Telemetry Configuration
   # Allows configuring telemetry settings.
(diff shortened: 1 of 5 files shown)


lib/rage/logger/logger.rb
```diff
@@ -132,6 +134,29 @@ def dynamic_context=(dynamic_context)
...
+        @filter_parameters = normalized_filters.freeze
+        @filter_parameters_matcher = Regexp.union(@filter_parameters)
+      else
+        @filter_parameters = nil
+        @filter_parameters_matcher = nil
+      end
+    else
+      @filter_parameters = nil
+      @filter_parameters_matcher = nil
+    end
```
rsamoilov: This method consists of the validation, forming the regexp, and calling `rebuild!`. Let's remove the validation/uniqueness/removing empty values completely from here - IMO it makes more sense to validate the parameters in the configuration layer. In Rage, configuration is "stateful" if it makes sense.

lib/rage/logger/logger.rb
```diff
@@ -196,7 +221,7 @@ def #{level_name}(msg = nil, context = nil)
         RUBY
       elsif @external_logger.is_a?(External::Static)
         # an object that implements Ruby's Logger interface is used as a logger
-        write_call = <<~RUBY
+        write_call = build_filtered_context_call(<<~RUBY)
```
rsamoilov: Nice! I didn't even know you could do that.

lib/rage/logger/logger.rb
```diff
@@ -233,7 +258,7 @@ def #{level_name}(msg = nil, context = nil)
           request_info: "Fiber[:__rage_logger_final].freeze"
         })
 
-        write_call = <<~RUBY
+        write_call = build_filtered_context_call(<<~RUBY)
```
rsamoilov: One thing I still haven't made up my mind about is whether we should be redacting context for [external loggers](https://rage-rb.dev/docs/logging/#external-loggers). Let me think about it a bit more. Feel free to share your opinion too.

lib/rage/logger/logger.rb
```diff
@@ -296,6 +321,42 @@ def build_formatter_call(level_name, level_val)
...
+    Fiber[:__rage_logger_context] = filter_value(context)
+    yield
+  ensure
+    Fiber[:__rage_logger_context] = context
+  end
+
+  def filter_value(value)
+    if value.is_a?(Hash)
+      value.each_with_object({}) do |(key, nested_value), filtered|
+        filtered[key] = if @filter_parameters_matcher.match?(key.to_s.downcase)
```
rsamoilov:
```suggestion
        filtered[key] = if @filter_parameters_matcher.match?(key.to_s)
```

Let's instead build the regexp as case insensitive.

lib/rage/configuration.rb
```diff
@@ -411,6 +430,18 @@ def validate_input!(obj)
     end
   end
 
+  class FilterParameters < LogContext
```
rsamoilov: It's nice thinking to reuse `LogContext` behaviour, but the filter keys are not log context objects. Even more, I don't think we need a separate configuration class at all - it's OK to validate the parameters directly inside `filter_parameters=`.

lib/rage/configuration.rb
```diff
@@ -256,6 +256,25 @@ def log_context
...
+  # Matching keys will have their values replaced with `"[FILTERED]"` before the log entry is written.
+  # @return [Rage::Configuration::FilterParameters]
+  #
+  # @example Filter common secrets from structured logs
+  #   Rage.configure do
+  #     config.filter_parameters = [:password, :token, :secret]
+  #   end
+  def filter_parameters
+    @filter_parameters ||= FilterParameters.new
+  end
```
rsamoilov: Is this method used anywhere? If not, it might make sense to remove it.

lib/rage/configuration.rb
```diff
@@ -256,6 +256,25 @@ def log_context
...
+
+  # Replace the current list of filtered log parameter keys.
+  # @param parameters [String, Symbol, Array<String, Symbol>, nil] one or more parameter filters
+  def filter_parameters=(parameters)
```
rsamoilov: Please don't hate me 😄  But I looked at your initial proposal one more time, and I think I like the "redact" terminology more - it's more precise, and distinguishing from Rails' `filter_parameters` makes sense considering this setting is about structured data in logs and has nothing to do with controller parameters specifically. What about `log_redact_keys=` instead of `filter_parameters=`? If we go this way, we should probably also update the placeholder from `[FILTERED]` to `[REDACTED]`.

lib/rage/configuration.rb
rsamoilov, line 1298: Let's move the assignment from below into here.

rsamoilov (commented):
Looks nice! Sorry for the delay - I left some comments.

lib/rage/logger/logger.rb
RedZapdos123: Moved validation, deduping, and empty-value cleanup into `Rage::Configuration` so `Rage::Logger` now only receives prepared keys, builds the matcher, and rebuilds.

lib/rage/logger/logger.rb
RedZapdos123: The matcher is now built as case insensitive so matching uses `key.to_s` directly.

lib/rage/configuration.rb
RedZapdos123: Removed the separate `FilterParameters` class and validate the redact keys in the configuration setter.

lib/rage/configuration.rb
RedZapdos123: That's fine. Renamed the public API to `config.log_redact_keys=`. I also changed the placeholder to `[REDACTED]`.

lib/rage/configuration.rb
RedZapdos123: Okay. `log_redact_keys=` now applies the keys to an existing logger immediately. `logger=` also reapplies stored keys when the logger is assigned later.

lib/rage/configuration.rb
RedZapdos123: No. So Removed the getter and kept this as a setter based configuration API.

lib/rage/logger/logger.rb
RedZapdos123: Wrapping the generated call path ended up being cleaner than I expected.

lib/rage/logger/logger.rb
RedZapdos123: I kept the redaction for external logger context for now because those backends still consume Rage's structured context payload so I thought `config.log_redact_keys=` should behave consistently across logger backends.

Else the same setting would redact built in logs but still leave the same keys unredacted in external logger context which felt interesting. `request_info` is still unchanged as discussed. If you will rather keep external logger payloads raw, I can scope this down to the built in logger paths only.

rsamoilov (approved):
Looks great!

Issue #379, opened 2026-08-03
[Telemetry] Observability: Metrics & Context Propagation


Abishekcs:
### Phase 1 Telemetry primitives (rage)
- [x] `Rage::Telemetry.every(ms)`, wraps `Iodine.run_every` handles boot timing internally (`Iodine.running?` and `on_state(:on_start)`)
- [x] Write Specs along with each PR. 

### Phase 2  Socket & connection metrics (rage & rage-iodine)
`rage-iodine` exposes no **accept-queue depth**. We add them, following the existing pattern (`task_inc!`/`task_dec!`/`stopping?` core fn in `fio.c`, declaration in `fio.h` and Ruby wrapper + registration in
`iodine_defer.c`).

Metric:

- [x] **Socket backlog (Accept Queue)**, new C method `Iodine.socket_backlog` in iodine: save the listener's fd when `fio_listen` runs, then read the kernel's accept-queue length (handshaked (ESTABLISHED) connections waiting for `accept()`) via `getsockopt(TCP_INFO)`   `tcpi_unacked`.
- [x] `Rage::Telemetry.queued_connections`, method-based reader exposing the Iodine server accept queue.
- [x]  Update existing spec or add new if required according to the changes.

### Phase 3 OpenTelemetry metrics (opentelemetry-instrumentation-rage)
- [ ] Event loop lag via `Telemetry.every`
- [ ] GC pressure gauges via periodic `GC.stat` sampling
- [ ] Observable gauges for all `Rage raw` keys
- [ ] Test the metrics are correct using the SDK's , then confirm they arrive at a real Collector running locally

### Phase 4 Grafana dashboard
- [ ] Connect to Grafana dashboard

### Phase 5 Context propagation (rage)

- [ ] Make `Current.user` (and all CurrentAttributes) survive every fiber boundary — enqueue, task execution and Fiber.schedule

### Phase 6 Benchmark & docs
- [ ] Benchmarks
- [ ] Complete spec coverage for the new metrics layer
- [ ] Docs + production monitoring guide

rsamoilov:
> Rage::Telemetry.raw, symbol-key read interface

It doesn’t really have to be a symbol-key read interface. Could regular methods work better here?

> :backlog_size from Rage::Deferred::Queue#backlog_size (already tracked)

`backlog_size` may not be the best name, but what I meant by it was the same thing you described as `Iodine.socket_backlog`. Exposing the deferred backlog size can also be useful but is not a priority. 

[...]

PR #396, opened 2026-09-03
Add module Rage::Telemetry::Capacity

Abishekcs:
#379 

### What this PR does?

Exposes the number of connections currently waiting in the kernel's accept queue for the server's listening sockets how many clients have connected but haven't been picked up by the app yet. Backed by Iodine.queued_connections, which walks Iodine's own listeners and reads each one's queue depth from the kernel.

Adds two specs: one confirms the reported count matches the number of connections actually parked in the queue, the other confirms the method works correctly when called from inside a .every block.

### Screen Recording

https://github.com/user-attachments/assets/c56b9176-31a3-49b6-aff2-258e6119ac0d

### DIFF
--- a/lib/rage/telemetry/telemetry.rb
+++ b/lib/rage/telemetry/telemetry.rb
@@ -36,6 +36,13 @@ module Rage::Telemetry
     __registry.keys
   end
 
+  # Returns the number of established connections currently waiting in the
+  # kernel's accept queue for the server's listening socket(s).
+  # @return [Integer] the accept-queue depth
+  def self.queued_connections
+    Iodine.queued_connections
+  end
+
   # Registers a block to be executed repeatedly at a fixed interval while the server is running.
   # The block is run inside a fiber, so blocking I/O inside it (e.g. flushing metrics to a
   # collector) will not block the server.
--- a/spec/telemetry/telemetry_spec.rb
+++ b/spec/telemetry/telemetry_spec.rb
@@ -56,6 +56,59 @@ RSpec.describe Rage::Telemetry do
     end
   end
 
+  describe ".queued_connections" do
+    before do
+      Fiber.set_scheduler(Rage::FiberScheduler.new)
+    end
+
+    after do
+      Fiber.set_scheduler(nil)
+    end
+
+    it "reports the accept-queue depth of the server's listeners" do
+      require "socket"
+
+      s = TCPServer.new("127.0.0.1", 0)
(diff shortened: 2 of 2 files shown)


spec/telemetry/telemetry_spec.rb
```diff
@@ -56,6 +56,59 @@
...
+      s = TCPServer.new("127.0.0.1", 0)
+      port = s.addr[1]
+      s.close
+      result = nil
+
+      Iodine.workers = 1
+      Iodine.on_state(:on_start) do
+        Iodine.listen(port: port, handler: -> { [200, {}, ["ok"]] })
+        socks = 5.times.map { TCPSocket.new("127.0.0.1", port) }
+        Iodine.run { result = described_class.queued_connections }
```
Abishekcs, line 80: @rsamoilov The test seems to fail. I suppose it is because the new changes made to Iodine are not yet reflected?

rsamoilov: Right, they haven't been released yet. Tagged you on https://github.com/rage-rb/iodine/pull/20.

rsamoilov: Published Iodine 5.6.0. The tests should work now.

lib/rage/telemetry/telemetry.rb
```diff
@@ -36,6 +36,13 @@ def self.available_spans
     __registry.keys
   end
 
+  # Returns the number of established connections currently waiting in the
+  # kernel's accept queue for the server's listening socket(s).
+  # @return [Integer] the accept-queue depth
+  def self.queued_connections
```
rsamoilov, line 42: Let's move this method into a new `Rage::Telemetry::Capacity` module.

lib/rage/telemetry/telemetry.rb
```diff
@@ -36,6 +36,13 @@ def self.available_spans
...
+    Iodine.queued_connections
```
rsamoilov, line 43:
```suggestion
    Iodine::Perf.queued_connections
```

rsamoilov, line 43: Also, let's maybe raise `NotImplementedError` if `queued_connections` returns `nil` - this will make it more explicit that the method is not supported on non-Linux platforms

Abishekcs:
@rsamoilov PR is ready for review.

rsamoilov (approved):
👍

